PR #10, opened 2023-10-19
feat(install): Arch user repository pkgbuilds // installation guide

pwnwriter:
I just packaged `presenterm` for [**aur**](https://aur.archlinux.org/). Any arch or based user can install it using aur helpers like yay/paru.


mfontanini:
Hey @pwnwriter, I just released version 0.3.0 and I'm not sure what to update here. e.g. what is the `sha512sums` in here, etc. Can you let me know/create a PR to fix those? Thanks!

PR #21, opened 2023-10-23
Use bat's syntect syntax set

### LINKED ISSUE #20: How I could achieve Elixir syntax highlighting?
Actually, How I could achieve syntax highlighting for a language that appears not to be supported officially in Sublime Text (syntect).

mfontanini:
This pulls in bat's syntect syntax set. This is a draft and needs this a bit nicely with a script to download it, and also pull in all other languages that are here but aren't there in syntect's default syntax set.

Fixes #20


mfontanini:
@dpc could you check if this breaks the nix flake? I've never used nix and I don't know how to verify that this doesn't break it. This is adding a new directory and a file is being pulled from it during compilation time. If you could add a github action to check for that on PRs or at least let me know how I can verify this doesn't break, that'd be great. Otherwise the flake will likely eventually break.

dpc:
```
> nix build github:mfontanini/presenterm
18:39:31 ~ took 2m20s 
```

Unless you add some new directories, it shouldn't require any maintenance. 

If you want I could add a workflow file to build in it in the CI.

mfontanini:
> Unless you add some new directories, it shouldn't require any maintenance.

Right that's what I did here, hence the question. So any new root directory should be added to that file and that's it? But yeah a workflow (assuming this doesn't take too long?) would be nice. No rush tho, I don't expect more directories popping up soon.

dpc:
As far as I can tell these new directories are not being used to build Rust. Do these files need to be present somewhere at runtime?

flake.nix
```diff
@@ -26,6 +26,7 @@
           ".cargo"
           "src"
           "themes"
+          "syntaxes"
```
dpc, line 29: **thought:** Oh, you've already added it? Then it's done. :D

mfontanini:
At build time. The file is being `include!`d at build time. And yep, I added it because I felt like I probably should but I couldn't verify that it actually worked. Sweet, thanks!

Issue #41, opened 2023-11-04
[feat] highlight code chunks


baggiponte:
Can't wait to try presenterm! I tried scouting for a feature that I love about `sli.dev`, the tool that I use for most of my presentations: [code highlight](https://sli.dev/guide/syntax.html#line-highlighting). Would you be interested in presenterm supporting it?

mfontanini:
That looks interesting, it would be nice to support it! Is there sunset of features you use more there? I think that should be fairly simple to implement except the `1|2|all` one as I think it needs something presenterm doesn't have (but I've been thinking of adding, that is the ability to pop content and replace it with something).

baggiponte:
> Is there sunset of features you use more there?

Not that I think of!

> except the `1|2|all` one

Do you think you could just display multiple slides - such as when you export to PDF a slideshow with elements that appear incrementally?

mfontanini:
Right, the problem is these are not really incremental changes. The highlighted code is changing (well, its formatting is) so it's not an additive change, it's a mutation over the existing contents of the slide. I'll look into this next though, I like this feature!

mfontanini:
Hey @baggiponte , I implemented support for:
* `+line_numbers` in a code block, which enables light numbers.
* `{1,2,5-10}` style highlighting, which only highlights lines 1,2, and 5 through 10.
* `{1|all|5-10}` style highlighting, which highlights line 1 first, then all of them, then 5 through 10.

Would you mind giving this a try and let me know if anything looks off? Thanks!

mfontanini:
This is an example where I move forward and back on this presentation:

[![asciicast](https://asciinema.org/a/dpXDXJoJRRX4mQ7V6LdR3rO2z.svg)](https://asciinema.org/a/dpXDXJoJRRX4mQ7V6LdR3rO2z)

PR #46, opened 2023-11-10
Allow enabling code block line numbers

mfontanini:
This allows enabling line numbers on code blocks by using the `+line_numbers` attribute.


mfontanini:
Relates to #41

PR #50, opened 2023-11-20
Feature: Load syntax highlighting themes from bat's config dir

rmartine-ias:
Hi! I was trying out your project, and wanted to be able to load my own syntax highlighting theme (`tokyonight_day`). [Bat](https://github.com/sharkdp/bat) also uses `syntect`, and has a [directory](https://github.com/sharkdp/bat/blob/7658334645936d2a956fb19aa96e6fca849cb754/src/bin/bat/directories.rs#L28) for custom themes, so I figured it would be neat if presenterm could pick up themes from there as well.

I haven't written much Rust before, so this is more of a starting point/suggestion than an actual PR -- there aren't any tests, this panics if the dir doesn't exist, it doesn't take into account the existence of the `BAT_CONFIG_DIR` variable, and it isn't the most forward-thinking solution (maybe this could be an env var `PRESENTERM_TMTHEMES_DIR` that one points at the bat themes dir instead?). That said, it works :)

Screenshot 2023-11-20 at 3 24 06 PM

Let me know if this is something you'd want to implement yourself, or if you'd prefer I take a proper crack at it. I can't promise any timeliness if I do.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,6 +489,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "etcetera"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "136d1b5283a1ab77bd9257427ffd09d8667ced0570b6f938942bc7568ed5b943"
+dependencies = [
+ "cfg-if",
+ "home",
+ "windows-sys 0.48.0",
+]
+
+[[package]]
 name = "exr"
 version = "1.71.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -592,6 +603,15 @@ name = "hex"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
+
+[[package]]
+name = "home"
+version = "0.5.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5444c27eef6923071f7ebcc33e3444508466a76f7a2b93da00ed6e19f30c1ddb"
+dependencies = [
+ "windows-sys 0.48.0",
+]
 
 [[package]]
 name = "iana-time-zone"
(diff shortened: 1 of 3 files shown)


mfontanini:
Thanks for the report/change! presenterm already uses `bat`'s syntaxes for code highlighting (e.g. to support more languages than syntect's defaults) so I think it would make sense to also embed their code highlighting themes as they've done a nice job with that. This should be a simple change but it involves moving a couple of things around so I'll give it a shot now.

mfontanini:
Phew this opened a pandora's box; I wasn't giving proper acknowledgements to `bat` and the syntaxes/themes it uses. I fixed that and started with this change but I didn't get there. There's 2 parts here and I only addressed one of them:
* One is it would be nice to pull in all of bat's themes because there's a bunch more than the few that syntect supports by default. This I have almost done, I'll try to finish it tomorrow.
* The other one is being able to pull in custom themes. This needs a bit of tweaking as the theme (presenterm's theme) needs to be able to set it so it needs to be configurable.

mfontanini:
Well this is a bit hellish. So the default themes in syntect use rgb colors but the ones supported by bat use ansi colors. This looks good in bat because they have their own [post processing](https://github.com/sharkdp/bat/blob/master/src/terminal.rs#L6) of themes to make these look nice. I need to dig into this a bit more; all code highlighted using these themes look bad even after trying some of that post processing.

rmartine-ias:
Oh wow, thanks for working on this! I knew `bat` was doing something extra (it has to build the theme cache) but didn't realize there was additional post-processing. It would be amazing to have the rest of the themes by default.

mfontanini:
I just merged a PR that adds support for all of bat's themes. Do you still feel the need to have your own or are bat's ones good enough?

rmartine-ias:
Thank you for adding all these! The theme I really want is [Tokyonight Day](https://github.com/folke/tokyonight.nvim/blob/main/extras/sublime/tokyonight_day.tmTheme), which `bat` doesn't come with by default, unfortunately.

mfontanini:
Alright, after #71 you can now drop the `.tmTheme` file into `~/.config/presenterm/themes/highlighting/` and that will make them available. e.g. then configure your theme to use `tokyonight-day` and it should work. Let me know otherwise!

rmartine-ias:
Works beautifully, thank you so much!

PR #63, opened 2023-11-24
Update information on using the Nix package

MikaelFangel:
I've updated the Nix section, so it explains the different options for Nix user, while still retaining the information on how to use the repo's own flake file. I don't know how familiar you're with Nix, so thought it would be easier, I proposed an addition to the README.md.


mfontanini:
Awesome, thanks!

> I don't know how familiar you're with Nix

Not at all so I appreciate the PR :pray:

PR #124, opened 2024-01-09
chore(entry): Don't panic if path isn't specified rather show helpful msg

pwnwriter:
If no path is provided, the current setup panics. Prooly showing some helpful msg is better ig? 

### Before: 

screenshot_2024-01-09_12-02-47

### After

screenshot_2024-01-09_12-03-22


pwnwriter:
Seems like I misunderstood; I'll create a new pr. Woopsie

mfontanini:
This looked mostly good to me, not sure why you closed it. It should probably use the nicer clap output like [this](https://github.com/mfontanini/presenterm/blob/master/src/main.rs#L110-L113) so the error feels like it comes from clap.

PR #126, opened 2024-01-10
Add hashes for windows artifacts

aliesbelik:
While trying newly created manifest for scoop (ScoopInstaller/Main#5409), noticed that hash files are absent for both generated Windows artifacts.
Suppose it could be useful to fix this injustice since generated hashes are already available for all the other artifacts.

Tested [here](https://github.com/aliesbelik/presenterm/actions/runs/7470080963/job/20328289340), sample release [here](https://github.com/aliesbelik/presenterm/releases/tag/v0.4.2).


mfontanini:
Awesome, thanks! I'll add these to the existing 0.4.1 release in a bit.

PR #132, opened 2024-01-12
Allow defining custom keybindings in config file

mfontanini:
This allows defining custom keybindings in the config file under `~/.config/presenterm/config.yaml`. I will document this along with the next release but you can do things like:

```yaml
bindings:
  next_slide: ["l", "<right>"]
  previous_slide: ["h", "<left>"]
  first_slide: ["1"]
  go_to_slide: [":<number><cr>"]
```

By default all mappings are the same as before, but they can be overridden one by one.

This for now performs only some checks on conflicting keybindings but it won't detect things like "potato" vs "ta" in its current form.

### DIFF
--- a/src/custom.rs
+++ b/src/custom.rs
@@ -1,6 +1,8 @@
 use serde::Deserialize;
 use std::{fs, io, path::Path};
 
+use crate::input::user::KeyBinding;
+
 #[derive(Clone, Debug, Default, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Config {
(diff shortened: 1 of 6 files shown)


mfontanini:
cc @pwnwriter

pwnwriter:
Looks really neat. But just curious, most of the rust related tools follow toml config. It'd have been more Fabolous if presenterm support that syntax.

mfontanini:
presenterm already uses yaml for things like the front matter and themes. I don't want to introduce another dependency/format unless it adds value, which I don't think toml would given the limited config we currently have. If at some point yaml becomes a problem, this can be changed.

PR #200, opened 2024-02-12
docs: add Scoop to package managers

nagromc:
Now that ScoopInstaller/Main#5409 is merged, we can add Scoop to the list of package managers.


mfontanini:
Awesome, thanks for creating the scoop packaging!

PR #202, opened 2024-02-14
Sort input file list

bmwiedemann:
Sort input file list
so that `themes.rs` builds in a reproducible way
in spite of non-deterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this is good.

Without this patch, `themes.rs` and the resulting `presenterm` binary would differ thusly:
```diff
-  theme_name: base16-eighties.dark
+  theme_name: GitHub
```

This patch was done while working on [reproducible builds for openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds).


mfontanini:
Sweet, thanks!

PR #211, opened 2024-02-17
Allow setting validate overflows mode in config file

mfontanini:
This allows configuring the overflow validation mode introduced in #209 in the config file via `defaults.validate_overflows`:
* `always` to always validate.
* `never` to never validate (this is the default).
* `when_presenting` to validate when running in presentation mode (`-p`).
* `when_developing` to validate when running in development mode (the default mode).

### DIFF
--- a/src/custom.rs
+++ b/src/custom.rs
@@ -55,16 +55,34 @@ pub struct DefaultsConfig {
 
     #[serde(default)]
     pub image_protocol: ImageProtocol,
+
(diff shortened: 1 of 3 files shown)


mfontanini:
Thanks for the heads up! Fixed in #712, see [docs](https://mfontanini.github.io/presenterm/configuration/settings.html#validate-terminal-overflows).

PR #228, opened 2024-03-23
feat(config): add JSON schema for validation,completion,documentation

mikavilpas:
# Help users write and maintain their configuration

This change introduces a JSON schema for the configuration file. This schema is used to validate the configuration file and provide feedback to the user if the configuration file is invalid.

https://github.com/mfontanini/presenterm/assets/300791/9b9e7c45-6f2f-4a45-9806-60a8413c2185

To use this schema, the schema must be referenced in the yaml configuration file like this:

```yaml
# yaml-language-server: $schema=./schema.json
```

> See the documentation for this feature here: https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#using-inlined-schema

This configures the [yaml-language-server](https://github.com/redhat-developer/yaml-language-server) to use the schema.json file to validate the configuration file. The language server is well supported in editors such as

- Neovim with [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#yamlls). Personally I also use [mason-lspconfig.nvim](https://github.com/williamboman/mason-lspconfig.nvim) to manage my language servers
- Visual Studio Code with the [YAML extension](https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml)
- More supported clients are listed in https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#clients

## Supported features

- Autocompletion of keys in the configuration file
- Hover support for keys in the configuration file
- Validation of the configuration file
  - Invalid keys are detected
  - Invalid values are detected for values that are supposed to be of a certain type or numeric range
  - Bundled in themes are suggested but any theme name is accepted (in case the user wants to use a custom theme)
- the top sections of the config have a markdown link to the documentation when hovered

## Testing the feature

To test this feature, you can do the following steps:

- fetch this git branch for testing
- open the configuration file in your editor
- add the magic comment to the top of the file with the path to the schema.json file
  - I used `# yaml-language-server: $schema=./schema.json` to develop this feature
- (restart your yaml-language-server if necessary)
- observe the autocompletion, hover and validation features by making edits in your editor

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,263 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Config",
+  "type": "object",
+  "properties": {
+    "bindings": {
+      "$ref": "#/definitions/KeyBindingsConfig"
+    },
+    "defaults": {
+      "description": "The default configuration for the presentation.",
+      "allOf": [
+        {
+          "$ref": "#/definitions/DefaultsConfig"
+        }
+      ]
+    },
+    "options": {
+      "$ref": "#/definitions/OptionsConfig"
+    },
+    "typst": {
+      "allOf": [
+        {
+          "$ref": "#/definitions/TypstConfig"
+        }
+      ],
+      "minimum": 1.0
+    }
+  },
+  "additionalProperties": false,
+  "definitions": {
+    "DefaultsConfig": {
+      "type": "object",
+      "properties": {
+        "image_protocol": {
+          "description": "The image protocol to use.",
+          "allOf": [
+            {
+              "$ref": "#/definitions/ImageProtocol"
+            }
+          ]
+        },
+        "terminal_font_size": {
+          "description": "Override the terminal font size when in windows or when using sixel.",
+          "default": 16,
+          "type": "integer",
+          "format": "uint8",
+          "minimum": 1.0
+        },
+        "theme": {
+          "description": "The theme to use by default in every presentation unless overridden.",
+          "type": [
+            "string",
+            "null"
+          ]
+        },
+        "validate_overflows": {
+          "description": "Validate that the presentation does not overflow the terminal screen.",
+          "allOf": [
+            {
+              "$ref": "#/definitions/ValidateOverflows"
+            }
+          ]
+        }
+      },
+      "additionalProperties": false
+    },
+    "ImageProtocol": {
+      "oneOf": [
+        {
+          "description": "Automatically detect the best image protocol to use.",
+          "type": "string",
+          "enum": [
+            "auto"
+          ]
+        },
+        {
+          "description": "Use the iTerm2 image protocol.",
+          "type": "string",
+          "enum": [
+            "iterm2"
+          ]
+        },
+        {
+          "description": "Use the kitty protocol in \"local\" mode, meaning both presenterm and the terminal run in the same host and can share the filesystem to communicate.",
+          "type": "string",
+          "enum": [
+            "kitty-local"
+          ]
+        },
+        {
+          "description": "Use the kitty protocol in \"remote\" mode, meaning presenterm and the terminal run in different hosts and therefore can only communicate via terminal escape codes.",
+          "type": "string",
+          "enum": [
+            "kitty-remote"
+          ]
+        },
+        {
+          "description": "Use the sixel protocol. Note that this requires compiling presenterm using the --features sixel flag.",
+          "type": "string",
+          "enum": [
+            "sixel"
+          ]
+        },
+        {
+          "description": "The default image protocol to use when no other is specified.",
+          "type": "string",
+          "enum": [
+            "ascii-blocks"
+          ]
+        }
+      ]
+    },
+    "KeyBinding": {
+      "type": "string"
+    },
+    "KeyBindingsConfig": {
+      "type": "object",
+      "properties": {
+        "close_modal": {
+          "description": "The key binding to close the currently open modal.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/KeyBinding"
+          }
+        },
+        "execute_code": {
+          "description": "The key binding to execute a piece of shell code.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/KeyBinding"
+          }
+        },
+        "exit": {
+          "description": "The key binding to close the application.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/KeyBinding"
+          }
+        },
+        "first_slide": {
+          "description": "The key binding to jump to the first slide.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/KeyBinding"
+          }
+        },
+        "go_to_slide": {
+          "description": "The key binding to jump to a specific slide.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/KeyBinding"
+          }
+        },
+        "last_slide": {
+          "description": "The key binding to jump to the last slide.",
+          "type": "array",
+          "items": {
+            "$ref": "#/definitions/KeyBinding"
+          }
+        },
+        "next": {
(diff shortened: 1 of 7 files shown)


mikavilpas:
Just realized one setting defaulted to a different value in the schema than what the actual default is. This is now fixed.

mfontanini:
Love the goal of this! I wonder if this could be dynamically generated by using something like [schemars](https://docs.rs/schemars/latest/schemars/). e.g. run `presenterm --generate-config-json-schema` or something and spit it to stdout.

My concern is this will eventually drift off from the config when new parameters are added and/or it is a bit harder for people to use given this is standalone file in the repo that they don't get if they download a binary artifact.

schema.json
```diff
@@ -0,0 +1,149 @@
...
+          "type": "number",
+          "minimum": 1,
+          "default": 16,
+          "markdownDescription": "Override the terminal font size when in windows or when using sixel"
+        },
+        "theme": {
+          "anyOf": [
+            {
+              "type": "string",
+              "enum": [
```
mfontanini, line 20: As an example of what I mean by drifting off, this list can easily get out of sync with the list of supported themes.

mikavilpas:
Oh I see your point, and I think you are 100% correct.

I'll see if I can build the schema automatically based on the rust types 🤔  👍🏻

mikavilpas:
question: It looks like `AsciiBlocks` is the default image rendering protocol and it cannot be configured in the configuration explicitly. Is this correct? I suppose it makes sense, but wanted to make sure.

mfontanini:
It should be configurable. This works for me, as in, starting presenterm with this config will complain because I don't have sixel enabled during compile time:

```yaml
defaults:
  image_protocol: sixel
```

mikavilpas:
Ok, I have a working proof of concept available for review and feedback now.

This version is able to generate the schema using schemars. The schema seems to work and should now be easy to keep in sync with the code.

Open question: how should you think the build system should generate this schema?

- Right now I just used a POC that runs when the program itself runs, but of course this cannot stay.
- I am not sure if this is now a runtime dependency which might have an effect on the program's size. Ideally I think this could be totally left out at runtime, but I don't know if that actually happens as schemars works via adding attributes to types that are used in production 

Do you have an idea how things like this are usually managed in rust projects?

mfontanini:
> I am not sure if this is now a runtime dependency which might have an effect on the program's size

I wouldn't worry about this. I'm generally cautious about adding dependencies but this is only pulling a couple and I think it's for a good cause so I don't mind the extra binary size.

What I was thinking though, is not to have the schema file in the repo at all and let people generate it manually if they feel like using it. e.g. if you're interested in using a yaml lsp, you can run `presenterm --generate-config-file-scema > /some/path/i/like` and then use the yaml lsp directive in your config file to point to this. Alternatively, we could at least add a step in the CI that verifies that generating the file spits out an identical file to whatever schema file we ship in the repo. Thoughts? I don't know if it's worth shipping it in the repo if people can generate it on their own.

src/custom.rs
```diff
@@ -45,21 +53,49 @@ pub enum ConfigLoadError {
...
     pub validate_overflows: ValidateOverflows,
 }
 
+// allow the theme to be any string, but suggest the valid themes based on the themes
+fn theme_schema(_gen: &mut SchemaGenerator) -> Schema {
+    let mut schema = SchemaObject::default();
+    schema.metadata().description =
+        Some("The theme to use by default in every presentation unless overridden.".to_string());
+    schema.instance_type = Some(SingleOrVec::Vec(vec![schemars::schema::InstanceType::String]));
+    schema.enum_values = Some(vec![
```
mfontanini, line 85: I know you spent some time to get this enum to work, but I'm not sure if it's the right way to go. Because you can have your own custom themes, this end up being misleading to users. I see that you do allow an arbitrary string as well, but I think if I see my autocompletion show a list of things (e.g. the build in themes), I would always assume I _need_ to use a value from that list so I may get confused if I have a local custom theme and it's not showing up.

mikavilpas: Sure, we can leave it out too. I'll revert it back to just a string with no suggestions for the built in themes.

mikavilpas:
> I wouldn't worry about this. I'm generally cautious about adding dependencies but this is only pulling a couple and I think it's for a good cause so I don't mind the extra binary size.

Okay 👍🏻 
I was thinking it could also be a separate cargo project/script in the repo in order to clearly limit the dependencies out of the production build - but this is certainly simpler.

> 
> What I was thinking though, is not to have the schema file in the repo at all and let people generate it manually if they feel like using it. e.g. if you're interested in using a yaml lsp, you can run `presenterm --generate-config-file-scema > /some/path/i/like` and then use the yaml lsp directive in your config file to point to this. Alternatively, we could at least add a step in the CI that verifies that generating the file spits out an identical file to whatever schema file we ship in the repo. Thoughts? I don't know if it's worth shipping it in the repo if people can generate it on their own.

I would go for the second option (shipping a file), but I guess a switch for creating the schema might also be a good option for advanced users. It's almost no extra work to do it as you said you are fine with having the schema related dependencies being shipped.

I forgot to mention this, but here's my initial idea for how the schema could be built and used:
- when the schema generation works, the schema can be committed to the repo so it's clearly associated with a git version
- the example configuration could include the magic comment and point to the project's github url for the schema
  - e.g. `# yaml-language-server: $schema=https://github.com/mikavilpas/presenterm/blob/master/config-schema.json`
- there could be a build step in github that verifies the schema is up to date, and fail the build if it's not
- there could optionally be a build step to verify the example configuration conforms to the schema that the configuration includes

This is essentially the setup that the lazygit project uses for their configuration (see their docs https://github.com/jesseduffield/lazygit/blob/master/docs/Config.md?plain=1#L19)

mikavilpas:
Just added a command for creating the schema:
image

mfontanini:
Sorry, I've been busy and had no time to look into this.

> I forgot to mention this, but here's my initial idea for how the schema could be built and used:

All of that sounds great! We can add a step in the CI in a separate PR but I think all of that sounds great :ok_hand:

src/main.rs
```diff
@@ -31,6 +31,10 @@ struct Cli {
     #[clap(long, hide = true)]
     generate_pdf_metadata: bool,
 
+    /// Generate a JSON schema for the configuration file.
+    #[clap(long, hide = true)]
```
mfontanini, line 35: I don't know if I'd `hide = true` this, this should be visible for people doing a `--help`

src/custom.rs
```diff
@@ -45,17 +48,24 @@ pub enum ConfigLoadError {
     Invalid(#[from] serde_yaml::Error),
 }
 
-#[derive(Clone, Debug, Deserialize)]
+// TODO(JsonSchema): suggest the valid themes based on themes.presentation.theme_names() but allow
```
mfontanini, line 51: As I commented before, I think this isn't necessary (this may be a leftover comment!)

mfontanini (approved):
This is great!

src/main.rs
mikavilpas: Ah, nice catch. Probably a copy paste mistake, now fixed.

src/custom.rs
mikavilpas: Yep 😄 
fixed now.

mikavilpas:
Ok I think things are looking good now. Just rebased to the main branch and cleaned the commit history.

I'll see if I can work on the CI part next, but as far as I'm concerned, this PR is ready to be merged.

mfontanini (approved):

mfontanini:
Thanks again!

mfontanini:
I can do the CI part later today, no worries :ok_hand:

mfontanini:
I think I'm going to rename the `schema.json` file to `config-file-schema.json`. I imagine we may want a schema for the theme file as well so this future proofs this. Will ping you in the next PR.

PR #229, opened 2024-03-23
chore(config): include strict_front_matter_parsing in default config

mikavilpas:
Hi, noticed this setting is missing from the default config.

I'm not sure if you'd like to include all keys there by default. In case it's not necessary, this change can be discarded 🙂


mfontanini:
Yep, this was the idea, I just forgot to add this one here. Thanks!

PR #231, opened 2024-03-24
refactor: fix all clippy warnings

mikavilpas:
The warnings became visible in rust-analyzer. I could also see the warnings on the command line by executing

`cargo clippy --all-targets`

I fixed all the warnings using rust-analyzer's LSP code actions.

The full list of warnings:

<details><summary>Details</summary>
<p>

```rust
    Checking presenterm v0.7.0 (/Users/mikavilpas/git/presenterm)
warning: this expression creates a reference which is immediately dereferenced by the compiler
   --> src/input/user.rs:535:43
    |
535 |         let result = binding.match_events(&events);
    |                                           ^^^^^^^ help: change this to: `events`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
    = note: `#[warn(clippy::needless_borrow)]` on by default

warning: explicit call to `.into_iter()` in function argument accepting `IntoIterator`
   --> src/presentation.rs:260:86
    |
260 |         self.chunks.into_iter().flat_map(|chunk| chunk.operations.into_iter()).chain(self.footer.into_iter()).collect()
    |                                                                                      ^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing the `.into_iter()`: `self.footer`
    |
note: this parameter accepts any `IntoIterator`, so you don't need to call `.into_iter()`
   --> /private/tmp/rust-20240215-5649-scngkd/rustc-1.76.0-src/library/core/src/iter/traits/iterator.rs:524:12
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
    = note: `#[warn(clippy::useless_conversion)]` on by default

warning: useless conversion to the same type: `presentation::SlideChunk`
   --> src/presentation.rs:656:29
    |
656 |             Slide::new(vec![SlideChunk::from(SlideChunk::default()), SlideChunk::default()], vec![]),
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `SlideChunk::from()`: `SlideChunk::default()`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion

warning: useless conversion to the same type: `presentation::SlideChunk`
   --> src/presentation.rs:657:29
    |
657 |             Slide::new(vec![SlideChunk::from(SlideChunk::default()), SlideChunk::default()], vec![]),
[...]

PR #234, opened 2024-03-31
ci: cargo clippy checks all targets for errors

mikavilpas:
This is a continuation of https://github.com/mfontanini/presenterm/pull/231 and should help in keeping all targets linted in the future. That PR only fixed the current situation - this one should make sure things stay clean in the future as well.

### DIFF
--- a/.github/workflows/merge.yaml
+++ b/.github/workflows/merge.yaml
@@ -49,7 +49,7 @@ jobs:
         run: cargo fmt --all -- --check
 
       - name: Run cargo clippy
-        run: cargo clippy -- -D warnings
+        run: cargo clippy --all-targets -- -D warnings
 
   nix-flake:
     name: Validate nix flake


mfontanini:
It's actually complaining about an import with this change, can you address that one? Thanks!

mikavilpas:
Yes, I'm actually just looking into it. I don't seem to be able to reproduce the error locally.

I wonder if CI uses a different version of the toolchain or clippy, or something. I can of course remove the error manually, but it might be annoying to keep CI happy in the future if the errors can't be seen locally.

@mfontanini would you have time to check if `cargo clippy --all-targets -- -D warnings` reports any errors for you?

mfontanini:
I don't see it locally. The problem is the CI is using the latest version of rust (which is a bad idea) so it will complain about things you may not see locally. If you `rustup update` you should see it.

But also this should use a pinned version of rust in the CI, otherwise this can happen. I was going to do that at some point but never did, maybe it's time.

mfontanini:
Closing as these should be fixed.

PR #233, opened 2024-03-30
wip: need to render images within working directory

chardskarth:
I'm new to rust and I really like this project.

I'm trying to render images using typst so tempdir wont work in my usecase. 

Upside of my PR: 
- able to render image using:
```
    ``` typest +render
    image("imagehererelativetocurrentfile.png")
    ```
```

Downside:
- Creates temporary `typst` files in the current working directory.

PR #235, opened 2024-03-31
Allow using image paths in typst

mfontanini:
This allows using images in typst by passing in the `--root` parameter to specify the root path. All images must have an absolute path starting from the directory where the presentation is. So e.g. if you have a layout like

```
foo/presentation.md
foo/image1.png
```

In order to use `image1.png` inside typst in that `presentation.md` file you need to:

~~~markdown
```typst +render
#image("/image1.png")
```
~~~

Fixes #233


chardskarth:
This works. Hmm, should you update the docs as well? Regarding adding images with typst?

mfontanini:
I tend to only update docs once the change is released, otherwise this will confuse people using the latest released version.

PR #239, opened 2024-04-21
Allow explicitly disabling footer in certain slides

### LINKED ISSUE #238: [Feature request] Skipping footer on certain slides
Hi there @mfontanini and the rest of the community!

I just recently discovered presenterm and I'm very much loving it. Congrats on making such a nice tool. 

I've been playing around with the different possible layouts and the footer configuration, and I was wondering if it would make sense for other people to skip the footer in certain slides (mostly the first and the last). 

I love having the footer as a subtle indication of progress during the talk, but it's kind of noisy when you are waiting to start the talk (projecting the intro slide) or, at the very end, in the Thanks/Q&A portion (the outro slide). I'd say those would be the main use cases, but I can see, too, how a specific slide that is already crowded with text or an ASCII diagram may benefit from skipping the footer. 

IMHO, it would be enough not to render the footer, the calculated interpolations for the template-based footer will remain untouched (e.g. even if the footer is not shown in the intro slide, it will still show 2/n in the next slide). 

The implementation could be based on a command or any special mark/comment that seems fit. If only the intro/outro are considered relevant, a setting in the footer frontmatter may be enough (`skip_on_intro`/`skip_on_outro`, etc.)

What do you think?

mfontanini:
This allows:
* Skipping the footer in the intro slide by setting the theme's `intro_slide.footer` property to `false`.
* Skipping the footer in any slide by using a command `<!-- no_footer -->`.

Fixes #238

### DIFF
--- a/src/theme.rs
+++ b/src/theme.rs
@@ -266,6 +266,10 @@ pub(crate) struct IntroSlideStyle {
     /// The style of the author line.
     #[serde(default)]
     pub(crate) author: AuthorStyle,
+
+    /// Whether we want a footer in the intro slide.
+    #[serde(default)]
+    pub(crate) footer: Option<bool>,
 }
 
 /// A simple style.
--- a/src/media/kitty.rs
+++ b/src/media/kitty.rs
@@ -11,7 +11,6 @@ use std::{
     io::{self, BufReader, Write},
     path::{Path, PathBuf},
     sync::atomic::{AtomicU32, Ordering},
-    u8,
 };
 use tempfile::{tempdir, NamedTempFile, TempDir};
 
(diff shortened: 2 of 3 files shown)


src/theme.rs
```diff
@@ -266,6 +266,10 @@ pub(crate) struct IntroSlideStyle {
     /// The style of the author line.
     #[serde(default)]
     pub(crate) author: AuthorStyle,
+
+    /// Whether we want a footer in the intro slide.
+    #[serde(default)]
+    pub(crate) footer: Option<bool>,
```
aitor, line 272: I was going to suggest renaming this to `footer_on_intro` or something like that, to make it more obvious that we are not skipping the footer always, just on the intro slide, but I've seen that this will be parsed from a YAML section that is already called `intro_slide` so it's clear that the setting only applies to that slide. No comments then! xD

Issue #242, opened 2024-05-18
python + other languages execution is possible! 


AnirudhG07:
@mfontanini I was wonder if I could actually run python in the bash script for exec.
So here is what i came up with!

```bash+exec
```bash +exec
#!/bin/bash

# Run Python code directly
python -c """
for i in range(5):
    print(f'Hello, Mr. {i})
""" // the triple quotes are important
\```   // ignore the backslash
```

This is very much successful! 
The output goes like-

```
       ——————————————— [done] ———————————

        Hello Mr.0!
        Hello Mr.1!
        Hello Mr.2!
        Hello Mr.3!
        Hello Mr.4!
```
It is amazing!
Now I would like to contribute based on this idea to the demo.md
And if we could make this as-
```python
```python +exec
# python code goes here
\``` // again ignore the backslash
```   
Then it would be great!

AnirudhG07:
Not just python, we can execute more languages using that hack that you can run everything through bash.
Here is an overview
```bash
```bash +exec
#!/bin/bash

# Run Python code
python -c """
print('Hello from Python!')
"""

# Write and run Java code
cat > HelloWorld.java << EOF
public class HelloWorld {
    public static void main(String[] args) {
        System.out.println("Hello from Java!");
    }
}
EOF
javac HelloWorld.java
java HelloWorld

# Write and run JavaScript code
cat > hello.js << EOF
console.log('Hello from JavaScript!');
EOF
node hello.js

# Write and run Rust code
cat > hello.rs << EOF
fn main() {
    println!("Hello from Rust!");
}
EOF
rustc hello.rs
./hello

# Write and run Go code
cat > hello.go << EOF
package main
import "fmt"
func main() {
    fmt.Println("Hello from Go!")
}
EOF
go run hello.go

# Write and run C code
cat > hello.c << EOF
#include <stdio.h>
int main() {
    printf("Hello from C!\n");
    return 0;
}
EOF
gcc hello.c -o hello
./hello

# Write and run C++ code
cat > hello.cpp << EOF
#include <iostream>
int main() {
    std::cout << "Hello from C++!" << std::endl;
    return 0;
}
EOF
g++ hello.cpp -o hello
./hello

# Write and run C# code
cat > hello.cs << EOF
using System;
class Program {
    static void Main() {
        Console.WriteLine("Hello from C#!");
    }
}
EOF
mcs hello.cs
mono hello.exe
\```
```
I have tested and every single one of these work.
[...]

PR #243, opened 2024-05-18
added python + other languages execution 

AnirudhG07:
This is the PR with respect to the issue #242 
I have added the code in demo.md!

 I hope you can add this and we can work on the python execution and more.

### DIFF
--- a/examples/demo.md
+++ b/examples/demo.md
@@ -153,6 +153,24 @@ done
 
 <!-- end_slide -->
 
+Shell code execution
+---
+
+Run commands from the presentation and display their output dynamically.
+
+```bash +exec
+#!/bin/bash
+
+python -c """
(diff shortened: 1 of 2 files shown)


AnirudhG07:
I have added languages.md as mentioned in issue #243, With this we can cover a lot of languages. A list of which is also provided within the file.
If possible we can integrate this to get a simple.
```bash
```language +exec
. . .
/```
```
What do you think @mfontanini

AnirudhG07:
you can close this PR cause it doesnt seem to be in need now? @mfontanini

mfontanini:
Yeah I think this PR isn't necessary anymore. Thanks for the initial issue/PR as it was the catalyst to implement this properly! :pray:

Issue #247, opened 2024-05-20
Unfinished code execution reports as `[finished with error]`


dmackdev:
Firstly - thank you for building this fantastic project! I recently tried several other terminal presentation tools, and I found `presenterm` to be best by far. In particular, for its of out-of-the-box and non-blurred image support and how great the default themes and styling look. Well done!

I have encountered an issue with the bash code execution feature. On one of my machines, executing any simple bash script from within a presentation yields the `[finished with error]` message in the execution output, nearly every time. Very occasionally the message does not appear. The bash script executes successfully when I execute it directly in the terminal.

E.g. for the script
```
echo "hi"
```
In the presentation, after executing it, the output would be:
```
hi
[finished with error]
```

FWIW, on another machine, I very rarely see the `[finished with error]` message, but it still comes up occasionally. Both machines have similar setups - both use `iterm2` and `zsh`.

I did some debugging: I believe the child code execution process had not exited by the time the output from the process is extracted - [this](https://github.com/mfontanini/presenterm/blob/master/src/execute.rs#L103) `try_wait` was returning `Ok(None)`. I also made a change to pipe `stderr` and display that, but there was no `stderr` output.

So my suggestions/questions from this are:
1. We should modify the `ProcessReader::run` method to not update the status to `ProcessStatus::Failure` if the child process has not exited yet. (We could potentially use a new `ProcessStatus::NotComplete` variant?)
2. Could the `try_wait` be replaced by the blocking `wait` to ensure the process has completed? `ProcessReader::run` is called in a new thread.
3. I think seeing `stderr` output would be preferable to the `[finished with error]` message. Seeing the actual `stderr` output would also help debugging. Any thoughts on this?

I would be happy to make contributions.

Cheers!

dmackdev:
> Could the try_wait be replaced by the blocking wait to ensure the process has completed? ProcessReader::run is called in a new thread.

Actually using blocking `wait` (or `wait_with_output`) would mean the code execution output only appears in the presentation after _all_ the code has finished executing - the presentation would not show the output line by line as soon as it is written to `stdout`.

mfontanini:
Thanks for the report! I managed to reproduce consistently by doing

~~~markdown
```bash +exec
echo "hi"
exec 1<&- sleep 1
```
~~~

I've never seen this locally unless I do that ^ but I guess it depends on how the OS handles things. I can imagine the race condition between closing stdout and the process dying is there and shouldn't be too rare to hit.

[...]

PR #250, opened 2024-05-26
fix(code-exec): wait for code execution process to exit completely

dmackdev:
Fixes https://github.com/mfontanini/presenterm/issues/247.

This PR ensures we wait for the code execution process to completely exit, when determining its exit status. This way, we do not erroneously show the `[finished with error]` output when the process had not exited yet.


mfontanini:
Thanks!

Issue #251, opened 2024-05-26
feat: Show `stderr` output from code execution


dmackdev:
This issue is split from https://github.com/mfontanini/presenterm/issues/247.

It would be great to show `stderr` output from code execution within a presentation.

This would:
- Help presenters to debug code snippets when creating a presentation,
- Allow presenters to deliberately show `stderr` output  within a presentation.

Two main approaches were discussed in https://github.com/mfontanini/presenterm/issues/247:

1. Capture `stderr` and show it separately after `stdout`.
2. Interleave `stdout` and `stderr` output, to emulate exactly what you would see from executing the code in the terminal.

Regardless of the solution, we must still see the output interactively, i.e. we should see each line of output immediately as it happens, instead of all the output at once after code execute process exits. This maintains the current functionality, as can be seen in the `bash` snippet in [examples/demo.md](https://github.com/mfontanini/presenterm/blob/master/examples/demo.md?plain=1#L146-L152) presentation, which has interleaved `echo`/`sleep` commands.

PR #252, opened 2024-05-26
feat: show `stderr` output from code execution

dmackdev:
Closes https://github.com/mfontanini/presenterm/issues/251.

This uses the [os_pipe](https://crates.io/crates/os_pipe) crate to capture both `stdout` and `stderr` from code execution, displaying them interleaved as you would see it in the terminal. The interactivity of seeing each output line as it happens (instead of all at the end) is retained.

https://github.com/dmackdev/presenterm/assets/79006698/0b93e4b3-96dd-4e98-9ac0-e03dd1b43514


mfontanini:
Awesome, thanks for the fix!

PR #253, opened 2024-05-26
feat: allow code snippet execution in any language

mfontanini:
This adds the groundwork to be able to allow code execution for any supported language rather than just bash scripts by using the `+exec` switch. The idea is that we files in the repo's `./executors` directory where the file name should be `<CodeLanguage>.sh` and should contain a shell script that indicates how that programming language is executed. The script will be invoked when a code snippet for that language is executed, taking as an argument the path to a file that contains the code snippet.

For interpreted languages like python the executor may be a script that contains a single line invoking the interpreter. For compiled languages it may contain an invocation to the compiler + one to the output binary.

Besides built in executors, you can define executors locally in `$config_path/executors` (`~/.config/presenterm` in linux) where each file should follow that naming format ^. 

There's still more work to be done to:
1. Add a lot more executors. This simply adds python as a showcase but adding more should now be trivial for contributors.
2. Allow defining executors within the presentation's path so that you can have a self contained reproducible presentation.
3. Allow processing output in real time as you now only see output when `python` finishes. I think something is using buffered output and is causing us to only receive the script's output once it finishes.

Relates to #242


executors/Python.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+exec python "$1"
```
dmackdev, line 3: Looks like python buffers `stdout`/`stderr` output by default. Running it with the `-u` flag disables this.
```suggestion
exec python -u "$1"
```
I have tried this and can see the live output!

mfontanini: Awesome! Thanks for trying it out

PR #254, opened 2024-05-27
feat: support hidden code lines

dmackdev:


dmackdev:
From https://github.com/mfontanini/presenterm/issues/242#issuecomment-2133699531:
> @dmackdev I think it would be nice if we didn't use /// for every language but this was configurable instead (defaulting to nothing so we don't have to preemptively define it for every language we support highlighting for). /// is valid rust so that would mean I can no longer use doc comments in code snippets, and even worse it would be pretty confusing if you don't know about this, add a /// section and then it doesn't show up in the presentation.
> 
> I like the doc tests approach in rust better where # is used to hide lines, see [this](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#hiding-portions-of-the-example). Could this be made optional and language-defined so CodeLanguage has some fn hidden_line_prefix(&self) -> Option<&'static str> that can be used to configure this? If you'd like to create the PR and move the conversation there that'd be great.

Totally makes sense @mfontanini - I can make those changes.

So is the thinking that from `CodeLanguage::hidden_line_prefix`, we return `Some("#")` for `CodeLanguage::Rust` and `None` for all other languages for now, and we can add more language specific hidden code lines as we go?

mfontanini:
> Would we just decide on some sensible prefix in those cases, should the need arise?

Yeah I'd fix those as they come. I imagine in most cases, rust being an exception, the hidden line prefix would be an extension to the language's comment syntax. So e.g. for python/shell script it could be `##`, for java `///` (but I don't know if this is used for something specifically so I'm unsure). But I wouldn't commit to a default for all languages blindly just in case as it may have a meaning in some of them already.

src/markdown/elements.rs
```diff
@@ -247,6 +271,14 @@ impl CodeLanguage {
     pub(crate) fn supports_auto_render(&self) -> bool {
         matches!(self, Self::Latex | Self::Typst)
     }
+
+    pub(crate) fn hidden_line_prefix(&self) -> Option<&'static str> {
+        match self {
+            CodeLanguage::Rust => Some("# "),
+            CodeLanguage::Python | CodeLanguage::Shell(_) | CodeLanguage::Bash => Some("## "),
```
dmackdev, line 278: I wonder if there is a better prefix over to use for these languages since a single `#` denotes a comment in them, so if you wanted to show comments in the snippet, that could potentially be confusing?

examples/hidden_code_lines.md
```diff
@@ -0,0 +1,111 @@
+---
```
dmackdev, line 1: This presentation was primarily for demonstration purposes, let me know if you think I should remove/amend it @mfontanini.

mfontanini:
I'm starting to have second thoughts about this change, specifically from the security angle. With this change you could run a code snippet in a presentation and it could do something like an `rm -rf $HOME` or other malicious actions without you knowing ahead of time. I think this could make people not want to share presentations / run someone else's presentations. The same goes for the "local executors to the presentation file" that we mentioned in the other issue. If we let the user define this on a per presentation basis, we'd be giving them arbitrary code execution capabilities on whoever runs their presentations' machines.

Thoughts? Am I too paranoid?

dmackdev:
A good point. I think the risk of executing malicious code is present regardless of hidden code lines. All it would take is a user to execute code in a presentation without actually inspecting/understanding: the snippet as it is seen in the presentation, the markdown file, or any local executors. Even without hidden code lines and local executors, there has to be some level of trust that a shared/downloaded presentation does not contain malicious code snippets. Of course, hidden code lines/local executor scripts could elevate this risk - if users rely only on viewing the presentation output and do not inspect the sources.

Indeed, other tools like `patat` disclose a warning about executing arbitrary code from presentations.

But perhaps we could add some sort of safeguard around this for hidden code lines - maybe a CLI flag `--enable-hidden-code-lines`, so users have to explicitly take some action to actually hide the prefixed code lines in the presentation. Without the flag, the prefixed code lines would appear in the presentation as normal (without the prefix). Not sure if that would be too much overhead for users, but just a thought.

mfontanini:
> there has to be some level of trust that a shared/downloaded presentation does not contain malicious code snippets

Yeah, definitely, but hidden code lines make this a lot more dangerous.

> maybe a CLI flag --enable-hidden-code-lines, so users have to explicitly take some action to actually hide the prefixed code lines in the presentation

I agree, if this was implemented we'd need this. However, then it would become annoying for the user that's building a presentation. Now, we could put this as a config option in the config file but this ends up defeating the purpose with foreign presentations. I have mixed feelings about this :/

dmackdev:
Fair enough @mfontanini.
In the absence of any other ideas or a way forward to address the security concerns, feel free to close this PR 🙂

examples/hidden_code_lines.md
mfontanini: This is great but for now I'd like to remove it. I usually write docs about new features along when I'm about to release the next version, and I'll use some of these examples there. I'm planning on, besides having docs for this (and snippet code execution in general, touching the `examples/code.md` example presentation to include more cases like these ones.

src/markdown/elements.rs
mfontanini: How about using `///`? I only dislike that in rust because it has a meaning but for the rest of the languages that could work.

mfontanini (commented):
Hey @dmackdev, I'm reviving this! I recently added a configuration/CLI arg that you need to use to explicitly enable snippet code execution in #276. Since that was my main concern and it's now there, I'd like to get this PR in as it adds a lot of value.

I left a couple of comments and it looks like there's some conflicts as well.

src/markdown/elements.rs
```diff
@@ -185,6 +185,30 @@ pub(crate) struct Code {
...
+    pub(crate) fn executable_contents(&self) -> String {
+        if let Some(prefix) = self.language.hidden_line_prefix() {
+            self.contents
+                .lines()
+                .map(|line| {
+                    let mut line = format!("{}\n", line);
+                    if line.starts_with(prefix) {
+                        line.replace_range(..prefix.len(), "");
+                    }
+                    line
```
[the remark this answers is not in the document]
dmackdev: Nice one! I have made that change, but also updated it to use `fold` instead of `map` as per the Clippy suggestion.

src/markdown/elements.rs
dmackdev: That makes sense, can roll with `///`.

examples/hidden_code_lines.md
dmackdev: Removed.

mfontanini:
Thanks, this is great!

PR #256, opened 2024-05-29
feat: use a cleaner execution error UI

mfontanini:
This changes the error reporting when the execution of a code snippet fails so that instead of appending a "[finished with error]" string at the end, we display "finished with error" at the little status box in between the code snippet and the execution output. See screenshot for reference. @dmackdev thoughts on this? I found the current UI kind of annoying when running broken code snippets. Maybe as an addition to this, the "finished with error" could have a red (configurable) color so it's more obvious.

image

### DIFF
--- a/src/processing/execution.rs
+++ b/src/processing/execution.rs
@@ -1,9 +1,10 @@
 use crate::{
     execute::{CodeExecutor, ExecutionHandle, ExecutionState, ProcessStatus},
-    markdown::elements::Code,
+    markdown::elements::{Code, Text, TextBlock},
     presentation::{AsRenderOperations, PreformattedLine, RenderOnDemand, RenderOnDemandState, RenderOperation},
     render::properties::WindowSize,
-    style::Colors,
+    style::{Colors, TextStyle},
+    theme::ExecutionStatusBlockStyle,
 };
 use itertools::Itertools;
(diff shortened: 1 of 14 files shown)


src/processing/execution.rs
```diff
@@ -24,13 +24,21 @@ pub(crate) struct RunCodeOperation {
     default_colors: Colors,
     block_colors: Colors,
     inner: Rc<RefCell<RunCodeOperationInner>>,
+    state_description: RefCell<&'static str>,
```
dmackdev, line 27: Minor point, but I wonder if instead of this field, we could derive the description from the process' `ProcessState`? And derive/implement `Display` for `ProcessState` and just use that.
There might be a slight complication because I think the process `handle` is an `Option` whose inner value is removed when the process completes.

dmackdev (commented):
Looks great! A definite improvement.

mfontanini:
Now with styling (still need to adjust colors in all themes, and I'm not too happy with the orange used in this theme already):

[asciicast](https://asciinema.org/a/u1R4zkpVgQOD6sC1fz8H2RVPE)

src/processing/execution.rs
mfontanini: I added styling now and this becomes more complicated as a `String` / `str` won't do anymore.

PR #257, opened 2024-05-29
feat: add kotlin script executor

dmackdev:
Adds `Kotlin.sh` to execute Kotlin scripts - see https://kotlinlang.org/docs/command-line.html#run-scripts.

Using Kotlin scripts instead of compiling to a `jar` first and then running it with `java` means that the `main` function boilerplate does not need to be included in the snippet.


mfontanini:
Thanks!

PR #269, opened 2024-06-15
feat: add rust-script as a new code executor

ZhangHanDong:
Add [rust-script](https://github.com/fornwall/rust-script) as a new code executor to support Rust code snippets with third-party crates.


examples/demo.md
```diff
@@ -153,6 +153,230 @@ done
 
 <!-- end_slide -->
 
+Shell Rust code execution 1
```
mfontanini, line 156: Can you remove this from the demo presentation? This is meant to be a very minimal presentation to showcase how the tool works only.

src/execute.rs
```diff
@@ -67,6 +67,11 @@ impl CodeExecutor {
                 let args: &[&str] = &[];
                 Self::execute_shell(interpreter, code.contents.as_bytes(), args)
             }
+            CodeLanguage::RustScript => {
```
mfontanini, line 70: This isn't needed, there's a wildcard match on any non `Shell` language below.

src/execute.rs
```diff
@@ -115,6 +120,7 @@ impl CodeExecutor {
         handle.program_path = Some(code_file);
         Ok(handle)
     }
+
```
mfontanini, line 123: Extra line here

executors/rust-script.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env rust-script
```
mfontanini, line 1: This should be `#!/usr/bin/env bash`, no?

mfontanini:
Thanks!

PR #270, opened 2024-06-20
feat: cache third party rendered images

mfontanini:
The mermaid snippet rendering introduced in #268 works but the mermaid cli is **very slow** so every modification to the presentation while developing is painfully slow. This adds caching at the third party image render layer. Images already live forever, this just uses a hashmap from the input string -> the image itself. We could also hash the inputs but that adds another dependency and inputs should be small enough that this would be negligible.

cc @mikavilpas


mikavilpas:
👍🏻👍🏻

PR #273, opened 2024-06-30
feat: render typst/latex/mermaid asynchronously

mfontanini:
This changes rendering of typst/latex/mermaid (now called "third party renders") to execute asynchronously. This wasn't needed for typst/latex as they're so fast you can't even tell it's rendering, but the mermaid CLI requires spinning up a browser (!!!) so it can take a couple of seconds to render. This causes the presentation experience, especially during development, to become very clunky as changing the presentation file just stalls presenterm for 2 seconds per chart until they're all rendered.

This changes that so that charts are rendered asynchronously by a pool of 2 threads (which I will make configurable in another PR). This now means you can:
* Open a presentation and scroll through slides while typst/latex/mermaid blocks are rendered and you'll see a "Loading..." text as a placeholder until they're loaded.
* Once loading finishes, the slide will automatically be updated to display the output image.
* Making a change on one of these blocks will make it render asynchronously so you'd see the same "Loading" text until it does.
* Because of the cache change on a previous PR, only the modified snippets are re-rendered.
* Snippets from slides closer to the beginning of the presentation are rendered first (e.g. they're all FIFO'd from first to last slide).
* Because renders can fail (e.g. mermaid syntax error) the "Loading.." can lead to an error. The error is shown the same way as before, except a bit better because it tells you which slide the error is in.

https://github.com/mfontanini/presenterm/assets/969090/9580bec8-9b6e-44b3-8920-09d7956176b1


mfontanini:
@mikavilpas I'd appreciate if you could give this a test and see if you find any quirks. The one thing I am aware of is that the errors, specifically for mermaid, are very long so lines overflow and they end up looking not so great (see the attached video, which is worse considering the window is tiny).

src/presentation.rs
```diff
@@ -569,18 +606,21 @@ pub(crate) trait AsRenderOperations: Debug + 'static {
...
 }
 
-/// A type that can be rendered on demand.
-pub(crate) trait RenderOnDemand: AsRenderOperations {
-    /// Start the on demand render for this operation.
+/// An operation that can be rendered asynchronously.
+pub(crate) trait RenderAsync: AsRenderOperations {
+    /// Start the render for this operation.
+    ///
+    /// Should return true iff the invocation triggered the rendering (aka if rendering wasn't
```
mikavilpas, line 613: Looks like a minor typo here (iff -> if)

mikavilpas:
Just tried this out - the difference is like night and day. Multithreading seems to work (2 graphs seem to get rendered in one go now). Caching works too, as does the loading indicator.

Great result, I'm really impressed!

mikavilpas (approved):

src/presentation.rs
mfontanini: [see](https://en.wikipedia.org/wiki/If_and_only_if) but this is probably the second or third time someone points this out so I should just stop using this :).

PR #275, opened 2024-07-01
Fix support nushell

PitiBouchon:
Fix the support for nushell on linux

_However nushell still doesn't work on windows because of the way the executor works_


mfontanini:
Thanks!

PR #296, opened 2024-07-18
feat: handle colored output

### LINKED ISSUE #272: Color output breaks line length
If a command produces colored output, it messes up the code box. It almost works, I think it's just counting the control sequence length incorrectly.

Screenshot 2024-06-29 at 12 52 37 AM

DrunkenToast:
Fix #272

Don't write a lot of Rust, so all feedback obviously welcome!

image

Colored output works, **however with the caveat of chunking not working**.
I am not sure how to approach chunking as it would be likely that I cut through ANSI escapes codes.
Keeping track of positions for where to slice seems messy.

Another expected issue is that the reset code clears out the codeblock format colors, not sure if we wanna resolve that too.
I was thinking that the rendering could intercept clear codes and apply the chosen colors but I think that is out of scope for this PR.

### DIFF
--- a/src/processing/execution.rs
+++ b/src/processing/execution.rs
@@ -10,12 +10,25 @@ use crate::{
 };
 use itertools::Itertools;
 use std::{cell::RefCell, mem, rc::Rc};
-use unicode_width::UnicodeWidthStr;
+
+#[derive(Debug)]
+struct OutputLine {
+    content: String,
+    /// Visible length of the output line, excluding ansi escape codes
+    length: u16,
+}
+
+impl OutputLine {
+    fn new(content: String) -> Self {
+        let stripped = strip_ansi_escapes::strip_str(&content);
+        Self { length: stripped.len() as u16, content }
+    }
+}
 
 #[derive(Debug)]
 struct RunSnippetOperationInner {
     handle: Option<ExecutionHandle>,
-    output_lines: Vec<String>,
+    output_lines: Vec<OutputLine>,
     state: RenderAsyncState,
     max_line_length: u16,
 }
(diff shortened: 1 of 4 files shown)


src/processing/execution.rs
```diff
@@ -102,11 +114,13 @@ impl AsRenderOperations for RunSnippetOperation {
 
         let block_length = self.block_length.max(inner.max_line_length.saturating_add(1));
         for line in &inner.output_lines {
-            let chunks = line.chars().chunks(dimensions.columns as usize);
-            for chunk in &chunks {
-                operations.push(self.render_line(chunk.collect(), block_length));
-                operations.push(RenderOperation::RenderLineBreak);
-            }
+            // TODO: I am not sure how to approach this properly as I would be cutting through
```
mfontanini, line 117: This is a good point I missed and will make this change a bit different I think. I'd suggest using this crate instead https://docs.rs/ansi-parser/latest/ansi_parser/ and parse each line with it and convert it into something that preserves the styles. For example, this could probably convert this into `TextBlock`s which you can then turn into render operations when you want to display them. Then, when you want turn that into `RenderOperation`s here, you would take the `dimensions.columns` as the max chunk size and if any line is bigger than you need to chunk it.

If this is more rust than you're comfortable (and dealing with the codebase, I'm actually not 100% sure this would work without changing anything about rendering strings) with, let me know and I get have a look. I'm starting to carve out a new release and I'd like this to be the last thing going in unless some issue comes up but I won't have time to work on this/the release until next week so no rush.

DrunkenToast: Yeah, was a bummer! I didn't know it got chunked. I'll see what I can do, I have a few ideas for something that can maybe work but no promises as I have quite a few other things to do as well.
But seems like a fun challenge if it doesn't get too difficult.

mfontanini:
This is great! Sorry I haven't had time to look into this, I found a bunch of bugs and have been prioritizing that. The only thing I'll add is the logic for splitting text and fit it into the screen is there already but you need to turn text into `WeightedTextBlock` and render it via `BlockLineText::Weighted` instead of `BlockLineText::Preformatted` but that may create other issues.

mfontanini:
Hey @DrunkenToast, I just wanted to know I started looking into this. I had to change a few internal things to make this work as expected, but I have a somewhat working version based on yours that parses lines only once after reading them rather than every time they're rendered. Thanks for setting the base for this change!

DrunkenToast:
> Hey @DrunkenToast, I just wanted to know I started looking into this. I had to change a few internal things to make this work as expected, but I have a somewhat working version based on yours that parses lines only once after reading them rather than every time they're rendered. Thanks for setting the base for this change!

Awesome! Thanks for the hard work ❤️

DrunkenToast:
closed in favor of #316

PR #314, opened 2024-07-29
feat: add mermaid docs and sensible defaults

calebdw:
Hello!

Changes:
- add mermaid theme docs
- change default mermaid bg to `transparent` (this seems to be how Github renders mermaid charts)
- add sensible mermaid default based on theme

I noticed that the `mermaid` theme parameters didn't make it into the documentation so I decided to add it.

I also tried to add sensible mermaid theme defaults to better match the supported themes instead of always having to override the theme. For instance, on the `dark` theme the original mermaid defaults (bg: white, theme: default) are quite jarring:

image

But updating to (bg: transparent, theme: dark) looks much nicer:

image

Thanks!

### DIFF
--- a/themes/catppuccin-frappe.yaml
+++ b/themes/catppuccin-frappe.yaml
@@ -100,3 +100,7 @@ footer:
 modals:
   selection_colors:
     foreground: "85c1dc"
+
+mermaid:
+  background: transparent
+  theme: dark
--- a/themes/catppuccin-macchiato.yaml
+++ b/themes/catppuccin-macchiato.yaml
@@ -100,3 +100,7 @@ footer:
 modals:
   selection_colors:
     foreground: "7dc4e4"
+
+mermaid:
+  background: transparent
+  theme: dark
--- a/themes/catppuccin-mocha.yaml
+++ b/themes/catppuccin-mocha.yaml
@@ -100,3 +100,7 @@ footer:
 modals:
   selection_colors:
     foreground: "74c7ec"
+
+mermaid:
+  background: transparent
+  theme: dark
--- a/themes/tokyonight-storm.yaml
+++ b/themes/tokyonight-storm.yaml
@@ -101,3 +101,7 @@ footer:
 modals:
   selection_colors:
     foreground: "e0af68"
+
+mermaid:
+  background: transparent
+  theme: dark
--- a/docs/src/guides/mermaid.md
+++ b/docs/src/guides/mermaid.md
@@ -35,3 +35,10 @@ sequenceDiagram
 It is recommended to change the `mermaid.scale` parameter until images look big enough and then adjust on an image by 
 image case if necessary using the `+width` attribute. Otherwise, using a small scale and then scaling via `+width` may 
 cause the image to become blurry.
+
+## Theme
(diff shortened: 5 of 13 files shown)


themes/catppuccin-frappe.yaml
```diff
@@ -100,3 +100,7 @@ footer:
 modals:
   selection_colors:
     foreground: "85c1dc"
+
+mermaid:
+  background: transparent
+  theme: neutral
```
mfontanini, line 106: This should be `dark`.

themes/catppuccin-macchiato.yaml
```diff
@@ -100,3 +100,7 @@ footer:
 modals:
   selection_colors:
     foreground: "7dc4e4"
+
+mermaid:
+  background: transparent
+  theme: neutral
```
mfontanini, line 106: This should be `dark`.

themes/catppuccin-mocha.yaml
```diff
@@ -100,3 +100,7 @@ footer:
 modals:
   selection_colors:
     foreground: "74c7ec"
+
+mermaid:
+  background: transparent
+  theme: neutral
```
mfontanini, line 106: This should be `dark`.

themes/tokyonight-storm.yaml
```diff
@@ -101,3 +101,7 @@ footer:
 modals:
   selection_colors:
     foreground: "e0af68"
+
+mermaid:
+  background: transparent
+  theme: neutral
```
mfontanini, line 107: This should be `dark`.

mfontanini (commented):
Thanks! Just a couple of changes on some darker themes that need `dark`. Feel free to double check them.

I actually tested this while adding mermaid support but somehow it looked like crap for me. I definitely was using the wrong parameter combo because it looks so much better this way. Thanks!

calebdw:
@mfontanini, I used `neutral` on the lighter "dark" themes because they seemed to not cause as much contrast:

Here is `neutral`:

image

And here is `dark`
image


But I can update all of them

mfontanini:
Hmm can you try a `sequenceDiagram`? The text in between the participants was showing up too light for me

calebdw:
Sure!

Here's the `neutral`:

image


and here's the `dark`:

image


Thanks for mentioning, `dark` does seem to better choice here :+1:

mfontanini:
Hmm now that I remember... When I was testing this I was trying it with pie charts. It's possible those look like crap in dark mode, I'm not sure. I'm away from computer now so I can't test those.

calebdw:
So here's `neutral`:

image


and here's `dark`:

image


I think `dark` is the best sensible default here, note that it's always possible to customize individual charts if desired/necessary:
~~~markdown
```mermaid +render
%%{init: {'theme':'forest'}}%%
  graph TD
    a --> b
```
~~~

mfontanini:
So weird, I wonder what exact combo I tried. Maybe I missed `dark` and only tried neutral. Anyhow, this looks great. Thanks!

calebdw:
Thank you sir!

Issue #289, opened 2024-07-15
Code blocks are not wrappable


DrunkenToast:
Long text is currently not supported with code blocks, they won't wrap properly as the characters are styled but it wraps after the styling is done.
Preferably it would be wrapped before the styling is applied.

![image](https://github.com/user-attachments/assets/4d84ce6e-97e7-438d-ac69-82f7592dc18a)

---

Expected outcome (mind the line numbers).

![image](https://github.com/user-attachments/assets/83c70e4f-e0ca-4c25-9e28-9722586d58f8)

---

I would understand a "Won't fix" label, since styling like this can be a whole can of worms, and ultimately the writer can simply split up the lines himself.

This would however make it easier to present on different screen sizes (especially considering you can create columns) and make line numbers a bit more convenient.

mfontanini:
This is something I'm aware of. Like you say, this could open up a can of worms. Also, in the way this is implemented, highlighting happens during presentation load time so it's not as easy as it sounds to do the wrapping because you only know where the end should be once the presentation is being displayed (as you need to know the window/rect size). I do see the issue though, so I'd like to have a look. It's just not as simple to fix as you'd imagine it should be unless things get moved around quite a bit.

DrunkenToast:
> so it's not as easy as it sounds to do the wrapping

Oh no, don't get me wrong, it sounds terribly difficult!
Working text is hell-ish!

Just wanted to open this issue so it's being tracked :)

PR #316, opened 2024-07-30
feat: process colored output from execution output

mfontanini:
This processes execution output's ansi escape codes and turns them into `WeightedTexBlock`. This can then be rendered so that we:
* Get colored output automatically.
* Handle lines that are too long to fit on the screen automatically.

This needs more testing, I just tested a couple of commands and it works fine but I didn't thoroughly test yet.

This is based on the initial work done on #296.

cc @DrunkenToast @calebdw I would appreciate if you could give this PR a test. It's only handling a very tiny subset of escape codes so I wouldn't be surprised if something failed. At least I tried basic things like `ls --color=always` and piping stuff to `lolcat` and they both seem to work fine to me.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -33,6 +33,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "ansi-parser"
+version = "0.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c43e7fd8284f025d0bd143c2855618ecdf697db55bde39211e5c9faec7669173"
+dependencies = [
+ "heapless",
+ "nom",
+]
+
+[[package]]
 name = "anstream"
 version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


DrunkenToast (commented):
Overall looks really good!
I did find an edgecase that doesn't behave like my shell:
image

Shell:
image

```bash
printf "\033[31;1;4mColors should get reset but not the underline:\n"

cowsay "Should get underlined but isn't" | lolcat -f
```

I don't immediately see a reason in this PR that would cause the issue, so I would assume it's somewhere in the renderer.

calebdw:
@mfontanini,

It looks pretty good to me as far as most use-cases go! I did notice it has trouble wrapping but looks like that's related to #289 

image

mfontanini:
Nice, thanks for testing it! Will fix both issues. 

> Codeblocks you can manually chunk up, but output of said codeblock is not as easy.

Don't worry, I won't merge until this is fixed. This should not look this way. Btw the changes that I'm adding here will allow simplifying a bunch of things + I _think_ will allow fixing the other issue you created re: splitting long code snippet lines.

mfontanini:
Alright, both issues are fixed:

image
image

calebdw:
That was fast!

mfontanini:
Merging, if anything comes out I can re-visit. There's lots of escape codes missing here so it's possible the output of some tools won't look right.

PR #317, opened 2024-07-30
feat: add event, location, and date frontmatter

### LINKED ISSUE #312: feat: add date to frontmatter
Hello!

Could you please add the date to the slide frontmatter that would be displayed on the intro slide? It's pretty standard to include the presentation date on the first slide. (Some other frontmatter options which might be useful are Event/Location information if the presentation is part of a large Conference/Summit/etc.).

I know this was asked for in #7 and that issue was closed because it's possible to add the date to the footer, but I'd rather not have to add the date to *every* slide.

As for the location, I'm not too picky as long as it  looks nice---but here's an idea:

image

Thanks!

calebdw:
Hello!

Closes #312 

This PR adds fields for the presentation `event`, `location`, and `date`---these are optional fields but are common frontmatter elements, particularly if the presentation is for a larger conference / event.

Example:
~~~markdown
---
title: How to Think Like a Designer
sub_title: It's easier than you think
event: Laracon US 2024
location: Dallas, TX
date: 2024-08-28
author: Jack McDade
---
~~~

image

Thanks!

P.S., I'm really new to Rust so sorry if I didn't do thing in the idiomatic Rust way!

### DIFF
--- a/src/presentation.rs
+++ b/src/presentation.rs
@@ -493,6 +493,18 @@ pub(crate) struct PresentationMetadata {
     #[serde(default)]
     pub(crate) sub_title: Option<String>,
 
+    /// The presentation event.
+    #[serde(default)]
+    pub(crate) event: Option<String>,
+
+    /// The presentation location.
+    #[serde(default)]
+    pub(crate) location: Option<String>,
+
+    /// The presentation date.
+    #[serde(default)]
+    pub(crate) date: Option<String>,
+
     /// The presentation author.
     #[serde(default)]
     pub(crate) author: Option<String>,
@@ -508,6 +520,19 @@ pub(crate) struct PresentationMetadata {
     /// The presentation's options.
     #[serde(default)]
     pub(crate) options: Option<OptionsConfig>,
+}
+
+impl PresentationMetadata {
+    /// Check if this presentation has frontmatter.
+    pub(crate) fn has_frontmatter(&self) -> bool {
+        self.title.is_some()
+            || self.sub_title.is_some()
+            || self.event.is_some()
+            || self.location.is_some()
+            || self.date.is_some()
+            || self.author.is_some()
+            || !self.authors.is_empty()
+    }
 }
 
 /// A presentation's theme metadata.
--- a/src/processing/builder.rs
+++ b/src/processing/builder.rs
@@ -283,11 +283,7 @@ impl<'a> PresentationBuilder<'a> {
         }
         self.footer_context.borrow_mut().author = metadata.author.clone().unwrap_or_default();
         self.set_theme(&metadata.theme)?;
-        if metadata.title.is_some()
-            || metadata.sub_title.is_some()
-            || metadata.author.is_some()
-            || !metadata.authors.is_empty()
-        {
+        if metadata.has_frontmatter() {
             self.push_slide_prelude();
             self.push_intro_slide(metadata);
         }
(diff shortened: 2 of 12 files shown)


docs/src/guides/themes.md
```diff
@@ -119,7 +119,7 @@ the right (but hey, you're free to do so!).
 The following elements support alignment:
 * Code blocks.
 * Slide titles.
-* The title, subtitle, and author elements in the intro slide.
+* The title, subtitle, event, location, date, and the author elements in the intro slide.
```
mfontanini, line 122: I appreciate the effort but I tend to not document new features until they're released. Otherwise people may try to use them using the latest version and be surprised they're not available. This could be set up in some form where you have the nightly/latest docs separately but it's not set up/don't know if it's worth it yet so I'd prefer to keep this undocumented until it's released.

examples/demo.md
```diff
@@ -12,6 +12,9 @@ An introduction slide can be defined by using a front matter at the beginning of
 ---
 title: My presentation title
 sub_title: An optional subtitle
+event: The event the presentation is for
```
mfontanini, line 15: Same here, this will cause the demo presentation to no longer work with the latest version so I'd rather keep it hidden until the next release.

mfontanini (commented):
This is great, thanks!

calebdw:
@mfontanini, the concern about documentation can be solved a few different ways---the best way is to version the docs so a user could see the docs for `master`, `v0.8`, `v0.7`, etc. by a version select on the docs site.

An easy fix in the near term though is to simply update `.github/workflows/docs.yaml` to only deploy new docs when a new version tag is pushed instead of on every push to `master` (I can easily include this change in this PR if you like).

I can remove the docs if you wish, but I do have several concerns about doing so:
1. It increases the maintenance burden at release time---it is much easier to update the docs with the PRs that introduced new features instead of trying to wait until ready to release and then put together all the docs
3. New features have the potential to be forgotten about and missed. The mermaid theme docs were missed and so I added them in #314
5. It is very difficult for nightly users to discover new features / options. They would need to parse through source code instead of just being able to read the docs on the `master` branch. This actually happened to me (I like to be on the bleeding edge of things), I only found out about the mermaid theme options because I was digging through the source code trying to figure out if it was possible to update the background color---it would have been much easier if the docs were at least updated on `master`

Just let me know how you'd like to move forward!

mfontanini:
> the best way is to version the docs so a user could see the docs for master, v0.8, v0.7, etc. by a version select on the docs site.

Yeah, I'm aware of this solution but I have limited time to spend on this project and this would be very low on my priority list. I'm sure however this would work it would be annoying though as I'll have to backport doc changes to all versions if it's a docs "fix" (e.g. a missing feature or something that's wrong) rather than a new feature being included.

> An easy fix in the near term though is to simply update .github/workflows/docs.yaml to only deploy new docs when a new version tag is pushed instead of on every push to master (I can easily include this change in this PR if you like).

Then if I find an issue or something missing in the current docs (e.g. your mermaid theme change), I'll need to manually remove all docs that only apply to nightly and not the latest release, generate the docs by hand, and push them which is a non negligible amount of work and very easy to mess up (e.g. end up including changes for nightly on the latest release's docs).

> It increases the maintenance burden at release time

I don't really mind this tho :). I can actually sit down and add/change/clean up everything all at once.

> New features have the potential to be forgotten about and missed

Fair but this is on me, I should have made a note to myself.

> It is very difficult for nightly users to discover new features / options

Fair, but I think in its current form the solution is pretty bad as well. I obviously don't have numbers on this but I imagine almost nobody (maybe you're the 0.1%) will clone the repo and manually generate the docs just to see what's new on nightly.

calebdw:
Alrighty :+1:, I removed the doc changes

mfontanini:
Thanks!

calebdw:
Thank you sir!

PR #318, opened 2024-07-31
feat: handle suspend signal (SIGTSTP)

### LINKED ISSUE #237: Handle suspending execution when receiving SIGTSTP signal (CTRL+Z)
This would improve presentations by being able to pause execution, do something in the terminal, and then resume execution at the same location the presentation was at previously.

Currently I either shift between two terminals or, kill the process and start it over again - navigating to where I was in the presentation.

calebdw:
Hello!

Closes #237 

This PR allows `presenterm` to be suspended / resumed with `<c-z>` and `fg`.

Thanks!

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -1,3 +1,4 @@
+---
 # yaml-language-server: $schema=https://raw.githubusercontent.com/mfontanini/presenterm/master/config-file-schema.json
 defaults:
   # override the terminal font size when in windows or when using sixel.
@@ -83,3 +84,6 @@ bindings:
 
   # the key binding to close the application.
   exit: ["<c-c>", "q"]
+
+  # the key binding to suspend the application.
+  suspend: ["<c-z>"]
--- a/src/presenter.rs
+++ b/src/presenter.rs
@@ -112,6 +112,10 @@ impl<'a> Presenter<'a> {
                 };
                 match self.apply_command(command) {
                     CommandSideEffect::Exit => return Ok(()),
+                    CommandSideEffect::Suspend => {
+                        self.suspend(&mut drawer);
+                        break;
+                    }
                     CommandSideEffect::Reload => {
                         self.try_reload(path, false);
                         break;
@@ -196,6 +200,7 @@ impl<'a> Presenter<'a> {
                 return CommandSideEffect::Reload;
             }
             Command::Exit => return CommandSideEffect::Exit,
+            Command::Suspend => return CommandSideEffect::Suspend,
             _ => (),
         };
         if matches!(command, Command::Redraw) {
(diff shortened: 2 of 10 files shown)


config.sample.yaml
```diff
@@ -82,4 +83,7 @@ bindings:
   close_modal: ["<esc>"]
 
   # the key binding to close the application.
-  exit: ["<c-c>"]
+  exit: ["<c-c>", "q"]
```
mfontanini, line 86: Yeah I'm not sure about this. vim uses `:q<enter>` which would be okay, but this is too easy to accidentally press as it is. You can always define this in your custom config file.

src/presenter.rs
```diff
@@ -348,10 +353,17 @@ impl<'a> Presenter<'a> {
             other => self.state = other,
         }
     }
+
+    fn suspend(&self, drawer: &mut TerminalDrawer<Stdout>) {
+        drawer.terminal.suspend();
+        unsafe { libc::raise(libc::SIGTSTP) };
```
mfontanini, line 359: I imagine this won't work on windows? If that's the case this needs a `#[cfg(...)]` guard so it's a no op in that platform.

mfontanini (commented):
Nice! Just a couple of comments.

config.sample.yaml
calebdw: @mfontanini, that's true and perhaps not the best example. But using just `q` is common enough in other programs---`lookatme` uses `q` and so does [`lazygit`](https://github.com/jesseduffield/lazygit) and others

src/presenter.rs
calebdw: Ah, good point!

src/presenter.rs
```diff
@@ -348,10 +353,20 @@ impl<'a> Presenter<'a> {
             other => self.state = other,
         }
     }
+
+    fn suspend(&self, drawer: &mut TerminalDrawer<Stdout>) {
```
mfontanini, line 357: This whole function should be guarded no? I would try to affect the windows behavior the least given it's the least tested platform

calebdw: Hmm, I'm not sure---if the whole function is guarded that means the method itself does not get compiled in right (so the function call above would fail)?

I don't use windoze so I can't test

calebdw: @mfontanini, if the function itself was guarded then you would also need to guard the function call since the function would not exist on windows:

```
error[E0599]: no method named `suspend` found for struct `Presenter` in the current scope
   --> src/presenter.rs:116:30
    |
43  | pub struct Presenter<'a> {
    | ------------------------ method `suspend` not found for this struct
...
116 |                         self.suspend(&mut drawer);
    |                              ^^^^^^^ method not found in `Presenter<'a>`
```

calebdw: Ah, that makes sense!

mfontanini:
This is great, thanks!

calebdw:
Thank you!

PR #319, opened 2024-07-31
fix: support mmdc on windows

bbawj:
I am not sure if Windows platform support is of any importance but I encountered an issue where `mmdc` cannot be found on Windows. This is due to this [issue](https://doc.rust-lang.org/std/process/struct.Command.html#implementations) where `Command::new()` on Windows works in this way: "if the file has a different extension, a filename including the extension needs to be provided, otherwise the file won’t be found.". 

Since mmdc on Windows is an alias for the `mmdc.cmd` script and not an `exe`, this updates the name used to find the mmdc tool.

### DIFF
--- a/src/tools.rs
+++ b/src/tools.rs
@@ -18,7 +18,12 @@ impl ThirdPartyTools {
     }
 
     pub(crate) fn mermaid(args: &[&str]) -> Tool {
-        Tool::new("mmdc", args)
+        let mmdc = if cfg!(windows) {
+            "mmdc.cmd"
+        } else {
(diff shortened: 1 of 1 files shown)


mfontanini:
> I am not sure if Windows platform support is of any importance

It is, I just didn't test this properly over there. I assumed since things like invoking cargo or other tools work this would work too but I guess not. Thanks!

mfontanini:
Looks like you need to run `cargo fmt` for the CI to pass.

mfontanini:
Thanks again!

PR #320, opened 2024-07-31
feat: handle code overflow

### LINKED ISSUE #289: Code blocks are not wrappable

mfontanini:
This makes it so that code blocks don't overflow to the right of the code block but instead how you'd expect them to: by creating a new line within the code block rect, properly colored and stuff.

image

Fixes #289

cc @DrunkenToast can you give this a shot? It seems to work well including line counts and all.

### DIFF
--- a/src/ansi.rs
+++ b/src/ansi.rs
@@ -23,7 +23,7 @@ impl AnsiSplitter {
             for p in line.ansi_parse() {
                 match p {
                     Output::TextBlock(text) => {
-                        self.current_line.0.push(Text::new(text, self.current_style.clone()));
+                        self.current_line.0.push(Text::new(text, self.current_style));
                     }
                     Output::Escape(s) => self.handle_escape(&s),
                 }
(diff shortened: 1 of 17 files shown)


DrunkenToast:
Woah, awesome! Didn't expect for this to be here so soon.

I did manage to break it however.
I am fairly sure this is simply because of the padding not being taken into account/incorrectly. Especially since it doesn't happen with output lines where there is no padding.

Examples:

Too small, goes outside box (2x padding):
image

Slightly too small, goes inside padding (1x padding):
image

Simply too much text (2x padding):
image

mfontanini:
Nice, thanks! I see the issue, I'll need to shuffle some things around. Weird that I resized the screen so many times and never saw this.

mfontanini:
Alright this is fixed. The only outstanding minor issue is when you shrink the window enough, the text will hit up to the code box's border. e.g. there won't be a 2 space padding at the end (see the the first `long long...` line in your first image). I can address this separately tho, as it's a smaller issue and this has been a bit of an annoying change.

PR #321, opened 2024-07-31
feat: allow closing with `q`

calebdw:
Hello!

This PR adds `q` to the list of exit commands---`q` is a common key used by many programs to exit such as:
- less
- man pages
- vim/neovim
- top/htop/btop
- lookatme
- lazygit
- etc.

This allows both `<c-c>` (the SIGINT signal) and `q` (the common exit key) to exit `presenterm`.

Thanks!


calebdw:
Thank you sir!

PR #323, opened 2024-07-31
fix: execute script relative to current working directory

### LINKED ISSUE #322: Running code from CWD
Currently code is ran from the /tmp directory, however this leads to unexpected results.

For instance a repository with slides might run scripts from that repository, however it will not be able to relatively refer to those scripts and will need to define the paths absolutely.

Now the problem arises when another person wants to run these slides, since the absolute paths would be incorrect.

I am not sure if this would cause any security issues, the slides are still explicit about what they do, they now just behave like any other executable and run from with the context of that cwd where the presentation was started.

Reasons why we would want this:
- Include other files in script block to focus on the important code, highlight in the codeblock
- Include libraries for languages such as Javascript
- Behaves as the writer/new users would expect

Not sure why we wouldn't want this but open to discuss.

---
```bash
pwd
```
`cd ~/slides`
`presenterm +x slides.md`

Expected behaviour:
`/home/<user>/slides`

Current behaviour:
`/tmp/<hash>`

DrunkenToast:
Intends to fix #322

I mostly just quickly wrote something so it would work for my use-case, so I didn't test all languages.

---

- [ ] Test all languages
- [x] Take presentation path instead of cwd

### DIFF
--- a/executors.yaml
+++ b/executors.yaml
@@ -5,12 +5,12 @@ bash:
 c++:
   filename: snippet.cpp
   commands:
-    - ["g++", "-std=c++20", "snippet.cpp", "-o", "$pwd/snippet"]
+    - ["g++", "-std=c++20", "$pwd/snippet.cpp", "-o", "$pwd/snippet"]
     - ["$pwd/snippet"]
 c:
   filename: snippet.c
   commands:
-    - ["gcc", "snippet.c", "-o", "$pwd/snippet"]
+    - ["gcc", "$pwd/snippet.c", "-o", "$pwd/snippet"]
     - ["$pwd/snippet"]
 fish:
   filename: script.fish
--- a/executors.yaml
+++ b/executors.yaml
@@ -1,7 +1,7 @@
 bash:
   filename: script.sh
   commands:
-    - ["bash", "script.sh"]
+    - ["bash", "$pwd/script.sh"]
 c++:
   filename: snippet.cpp
   commands:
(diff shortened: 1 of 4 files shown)


src/execute.rs
```diff
@@ -181,7 +182,7 @@ impl CommandsRunner {
         let child = process::Command::new(command)
             .args(args)
             .envs(env)
-            .current_dir(self.script_directory.path())
+            .current_dir(env::current_dir().unwrap())
```
mfontanini, line 185: I think the default may be this already, you may be able to avoid calling `.current_dir(...)` altogether. Can you give it a shot that way?

DrunkenToast: Derp, you're right. Removed it!

DrunkenToast:
Just a note for this PR:
This is only for the cwd of the process it runs from.

Bash works perfectly fine and as you would expect, however Javascript for example won't behave exactly as you'd expect.
Require statements will look relative to where the file/module is located, not from the current working directory. e.g. `process.cwd()` => works, `require('package')` => breaks, `require('./my-module')` => breaks.
And ultimately, this can be a thing with other languages/platforms as well, it's a bit out-of-scope for this project I think.

A workaround with node for this would be something like this `require(process.cwd() + '/node_modules/package')` or `require(process.cwd() + '/my-module')`. Which is still *way* better than what it was before but not pretty.

Personally I am content with this, since I mostly use bash but wanted to put this out there.

executors.yaml
```diff
@@ -1,7 +1,7 @@
 bash:
   filename: script.sh
   commands:
-    - ["bash", "script.sh"]
+    - ["bash", "$pwd/script.sh"]
 c++:
```
mfontanini, line 5: Looks like c and c++ still need to be updated

mfontanini:
> however Javascript for example won't behave exactly as you'd expect

I think this is no worse than it currently is., In fact it is still better because you can do _something_ to get the imports to work without hardcoding the presentation's path, which you couldn't before.

Thanks!

mfontanini:
Wait I think the current directory shouldn't be the default, it should be the base for where the presentation is. This assumes your cwd is the same as where the presentation is but that's not necessarily the case. This base oath is passed around, but not sure if it's accessible from here. We can merge this as is and I can look at it afterwards otherwise (still need the c and c++ changes before merging).

DrunkenToast:
> Wait I think the current directory shouldn't be the default, it should be the base for where the presentation is. This assumes your cwd is the same as where the presentation is but that's not necessarily the case. This base oath is passed around, but not sure if it's accessible from here. We can merge this as is and I can look at it afterwards otherwise (still need the c and c++ changes before merging).

Will look into it this evening.

As for the cwd being the presentation path: I was thinking about that too but a lot of applications behave like this so I didn't go through with it. But I think you're right, it would probably be best to take the path where the slides themselves are located.

DrunkenToast:
> Wait I think the current directory shouldn't be the default, it should be the base for where the presentation is. This assumes your cwd is the same as where the presentation is but that's not necessarily the case. This base oath is passed around, but not sure if it's accessible from here. We can merge this as is and I can look at it afterwards otherwise (still need the c and c++ changes before merging).

I resolved both of these.

mfontanini:
Looks like `cargo fmt` issues again. This actually requires running `cargo +nightly fmt` because there's a few nightly options enabled in rustfmt.toml, so this may be the issue.

calebdw:
@mfontanini, yes---I found that I had to specifically override the repo directory to use nightly. I'm not that familiar with Rust, but surely there can be a way to specify that nightly is required in the project configuration instead of everyone having to manually override?

mfontanini:
Not sure about that. There's a way to set the project to _build_ on nightly but you don't really want to enforce that, all you need is the `cargo fmt` to run on nightly. I've personally changed my editor to always use `cargo +nightly fmt` when formatting because it has no apparent drawbacks and ensures you get the nightly formatting rules whether it's enabled in a project or not.

DrunkenToast:
You can create a toolchain file: https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file
I wanted to create one in the beginning and submit a PR because I couldn't compile the project on stable rust but I was just one version behind.
If this project requires nightly, I highly recommend using a toolchain file.

mfontanini:
That's the thing, it doesn't require nightly to build. It needs nightly to be formatted. I've used a toolchain file but to specify it needs to build with it. I don't know if running `cargo fmt` will automatically use nightly if that's set in the toolchain file.

DrunkenToast:
Yeah sorry, misread it. From quick search online it indeed doesn't seem possible to only enforce fmt. But I do think the toolchain file makes all cargo/rust commands use that toolchain, would need to test it though.
You already have CI in place for it though :)

calebdw:
But why would you not just use the same toolchain to build with as well as format with? Seems it would just be easier that way to match?

PR #330, opened 2024-08-04
feat: allow snippets to be executed and replaced with output

### LINKED ISSUE #292: +render should work on all supported languages
The idea of adding `+render` is that is it rendered when the presentation is created.
It would be incredibly useful to have scripts that make up content of the presentation be rendered as well, however currently only latex and typst are supported.

For example CLI tools for creating text graphs can be put in a bash +render block, allowing you to insert CLI graphs easily in your presentation.
Even for programming languages, a quick and dirty js script in your presentation to programmatically create content in a slide could be handy.

This can also be a different option from `+render` if we want to preserve that for images.

mfontanini:
This adds the `+exec_replace` attribute that is basically an `+exec` except the output of the snippet's execution replaces the snippet itself. This allows using tools that emit ascii art, colorize outputs, etc.

I added a new set of configs to enable this (`-X` as a cli parameter and `snippet.exec_replace.enable` in the config file) as this adds a whole other level of danger when running someone else's presentation so I don't think the configs for `exec` are enough to protect users.

Fixes #292

### DIFF
--- a/config-file-schema.json
+++ b/config-file-schema.json
@@ -322,6 +322,14 @@
             }
           ]
         },
+        "exec_replace": {
+          "description": "The properties for snippet execution.",
+          "allOf": [
+            {
+              "$ref": "#/definitions/SnippetExecReplaceConfig"
+            }
+          ]
+        },
         "render": {
           "description": "The properties for snippet auto rendering.",
           "allOf": [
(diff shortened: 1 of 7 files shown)


calebdw:
> I added a new set of configs to enable this (-X as a cli parameter and snippet.exec_replace.enable in the config file) as this adds a whole other level of danger when running someone else's presentation

I'm not sure I see how exec + replace is any more dangerous than exec?

mfontanini:
Because it happens automatically on presentation load. So you open a presentation and this runs code without user intervention

DrunkenToast:
Good call with `-X`, fan of that.  
I think I like the replace better than the pre_process, although I am not sure if it is clear whether or not it render the replaced output as markdown or not.

mfontanini:
I'll merge this as is, if someone has a better name please shout

PR #332, opened 2024-08-09
feat: add php executor

calebdw:
Hello!

This adds the ability to execute PHP code.

Thanks!

### DIFF
--- a/src/processing/code.rs
+++ b/src/processing/code.rs
@@ -470,7 +470,9 @@ impl SnippetLanguage {
         use SnippetLanguage::*;
         match self {
             Rust => Some("# "),
-            Python | Bash | Fish | Shell | Zsh | Kotlin | Java | JavaScript | TypeScript | C | Cpp | Go => Some("/// "),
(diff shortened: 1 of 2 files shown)


src/processing/code.rs
```diff
@@ -470,8 +470,7 @@ impl SnippetLanguage {
         use SnippetLanguage::*;
         match self {
             Rust => Some("# "),
-            Python | Bash | Fish | Shell | Zsh | Kotlin | Java | JavaScript | TypeScript | C | Cpp | Go => Some("/// "),
-            _ => None,
+            _ => Some("/// "),
```
mfontanini, line 473: Please add `Php` here rather than changing the way this works. There's a [reason](https://mfontanini.github.io/presenterm/guides/code-highlight.html#hiding-code-lines) why this is the way it is.

calebdw: Updated, I get the reason but I figured since all the currently supported languages (except for Rust) use the `/// ` then an override would be added anyway if a language was added which did not support the triple slash.

mfontanini:
Thanks!

PR #334, opened 2024-08-12
feat: show link label and title

### LINKED ISSUE #325: Links lose their titles
Hello!

Currently, markdown links (`[foo](https://foo.bar)`) lose their titles and the actual link is shown italicized and underlined:

image

However, this removes context---oftentimes the link by itself doesn't tell me much. The title is there to explain the link or to hide away a lengthy url.

I have two ideas off the top of my head:

1. Show the *title* instead of the link, and open the link when the title is clicked on (would an actual link be possible in the terminal?)
2. If 1) is too complication / not doable, then show both, e.g., (the particular styling could be adjusted of course):
image

Thanks!

calebdw:
Closes #325 

Hello!

This PR renders the various link attributes (if they exist) to prevent lose of context/information. It's also nice to automatically style the relevant info so the user doesn't have to manually do it when copying markdown links from other location.

It turns out there's two other link attributes besides the url:
- **label** this is the portion between the  bracketss
- **title** this is the portion between the double quotes inside the link (this is most often shown as a tooltip when hovering over the link/label)

```markdown
* Links look like this [](https://example.com/)
* Links with titles look like this [](https://example.com/ "Example")
* Links with labels look like this [example](https://example.com/)
* Links with titles and labels look like this [example](https://example.com/ "Example")
```

image

I'm open to formatting suggestions, let me know what you think!

Thanks!


mfontanini:
Thanks!

calebdw:
Thank you!

PR #335, opened 2024-08-12
docs: document `all` keyword

calebdw:
Just adding some docs for this ability, thanks!

### DIFF
--- a/docs/src/guides/code-highlight.md
+++ b/docs/src/guides/code-highlight.md
@@ -90,6 +90,7 @@ Similar to the syntax used for selective highlighting, dynamic highlighting will
 code block are highlighted every time you move to the next/previous slide.
 
 This is achieved by using the separator `|` to indicate what sections of the code will be highlighted at a given time.
+You can also use `all` to highlight all lines for a particular frame.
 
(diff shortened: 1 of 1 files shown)


docs/src/guides/code-highlight.md
```diff
@@ -90,9 +90,10 @@ Similar to the syntax used for selective highlighting, dynamic highlighting will
 code block are highlighted every time you move to the next/previous slide.
 
 This is achieved by using the separator `|` to indicate what sections of the code will be highlighted at a given time.
+You can also use `all` to highlight all lines.
 
 ~~~markdown
-```rust {1,3|5-7}
+```rust {1,3|5-7|all}
```
mfontanini, line 96: The text that follows below is outdated after this change and there's an asciicast below that's outdated too. Maybe this should be added elsewhere so we don't need to change all of this

calebdw: How's this @mfontanini?

mfontanini: That works! I was going to tell you to update the paragraph and I'd update the asciicast later but I think just mentioning also works.

PR #338, opened 2024-08-15
feat: add template variables

calebdw:
Hello!

This PR adds some more variables that can be used in the footer template (basically everything available in the slide metadata):
- `{title}`
- `{sub_title}`
- `{location}`
- `{event}`
- `{date}`

There may be a better way of including all the metadata in the FooterContext without having to repeat all the definitions?

#### Default Template

I tried to add a default footer template, but I couldn't seem to decouple the footer template itself from the `template` style.

The following template is very common in presentations and I think it would make a great default:

```yaml
left: "{author}"
center: "{title}" # this can also be a short title
right: "{current_slide} / {total_slides}"
```

Ideally, I would like to just be able to set the style to `template` and have the default template used:
```yaml
theme:
  override:
    footer:
      style: template
```

However, it seems I must manually define the template for all of my slides

Thanks!

### DIFF
--- a/src/processing/builder.rs
+++ b/src/processing/builder.rs
@@ -290,7 +290,17 @@ impl<'a> PresentationBuilder<'a> {
         if let Some(options) = metadata.options.take() {
             self.options.merge(options);
         }
-        self.footer_context.borrow_mut().author = metadata.author.clone().unwrap_or_default();
+
+        {
+            let mut footer_context = self.footer_context.borrow_mut();
+            footer_context.title = metadata.title.clone().unwrap_or_default();
+            footer_context.sub_title = metadata.sub_title.clone().unwrap_or_default();
+            footer_context.location = metadata.location.clone().unwrap_or_default();
+            footer_context.event = metadata.event.clone().unwrap_or_default();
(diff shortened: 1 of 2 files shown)


src/theme.rs
```diff
@@ -475,9 +475,9 @@ pub(crate) enum FooterStyle {
 impl Default for FooterStyle {
     fn default() -> Self {
         Self::Template {
-            left: Some("{current_slide} / {total_slides}".to_string()),
-            center: None,
-            right: None,
+            left: Some("{author}".to_string()),
+            center: Some("{title}".to_string()),
```
mfontanini, line 479: This is too verbose of a default, I wouldn't change this at all. e.g. I don't think people should see my name in every slide + the title of the presentation can be too long and cause overflows.

If you want to use this as your default, you can set it globally for all your presentations https://mfontanini.github.io/presenterm/guides/configuration.html#default-theme. you can define a custom theme with these properties and use that as the default. This tool allows you to configure as much as possible so you can tweak it as much as you need without the tool enforcing anything on you

calebdw: I removed

mfontanini:
Thanks!

PR #339, opened 2024-08-15
feat: hide footer on intro slide by default

calebdw:
Hello!

This PR hides the footer on the intro slide by default which is fairly common in presentations and I think it would make a great default.

If you think differently then feel free to close :).

Note: maybe there's a better way of making this the default in the code without having to add it to all the themes? 

Thanks!

### DIFF
--- a/themes/catppuccin-frappe.yaml
+++ b/themes/catppuccin-frappe.yaml
@@ -68,6 +68,7 @@ intro_slide:
     colors:
       foreground: "b5bfe2"
     positioning: page_bottom
+  footer: false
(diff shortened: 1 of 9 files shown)


mfontanini:
I don't know about this one. I think this makes sense when using a `template` footer as it will likely have overlapping information with what the intro slide shows. However, all built-in themes default to the progress-bar style footer. For that one, I think not showing the footer in the intro slide actually looks weird and it also doesn't have the issue of overlapping information with the intro slide like the template one does.

PR #353, opened 2024-08-20
fix: don't wrap separator line

### LINKED ISSUE #343: Line-wrapping on executeable code block will cause the status line to wrap
If your code block is too big and wraps, the status line will take the original width and wrap as well, even though it could fit:
image
image

Expected behaviour is for the line to get shorter and not wrap.

mfontanini:
Fixes #343


calebdw:
You're cranking these out fast! 🚀

mfontanini:
There's nothing like starting the day at 5am with a little bit of bug fixing.

Issue #357, opened 2024-08-21
Option to remove or replace slide progressbar


dheerajshenoy:
I looked everywhere in the documentation if there was a way to interact with the slide progressbar shown at the bottom. It would be really nice if there was a way to either remove it completely or replace it or maybe have different kind of progress indicator, like current (slide number / total slides). I really love presenterm, thanks for creating this.

dheerajshenoy:
I found a section for this in the documentation [here](https://mfontanini.github.io/presenterm/guides/themes.html#template), the problem is solved i suppose. But then, if there was a flag like --no-progressbar or something like that, it would be cool.

mfontanini:
I turned the default footer into a template style footer in #358.

PR #358, opened 2024-08-21
feat!: use "template" footer in built-in themes

mfontanini:
This makes the footer in built in themes be a template with `{current_slide} / {total_slides}` in the right. The progress bar was "cool" but it's not very professional so I don't think it should be the default.

Relates to #339 #357


mfontanini:
cc @calebdw you'll probably like this :)

calebdw:
Cool beans!

PR #361, opened 2024-09-02
Add TOML highlighting. Resolves #360.

### LINKED ISSUE #360: Toml highlighting
I’d like to present some slides with toml in them. It appears toml isn’t currently supported for syntax highlighting. I’d be happy to take a look at adding support if you’d be happy with that? (I believe toml is supported by bat).

witchard:


mfontanini:
Thanks!

PR #366, opened 2024-09-09
feat: allow acquiring terminal when running snippets

### LINKED ISSUE #365: Resolving imports for Rust code execution
Hey again!

Is there a way to handle dependencies when using executable code blocks in the presentation. For example, I want to import a library like so:

````rust
```rust +exec
use ratatui::{
    crossterm::event::{self, Event, KeyCode, KeyEventKind},
    widgets::{Block, Paragraph},
};
```
````

But unfortunately this fails (when I do `C-e`) with import errors. 

Is this currently possible to do so with `presenterm`? If not, would you be interested in supporting this?

### PR DESCRIPTION
**Pull request by mfontanini**
This introduces a new `+acquire_terminal` snippet attribute which causes the snippet to take ownership of the terminal while the contents of it are being executed. This means the output won't be displayed inside the presentation but instead the presentation will pause, invoke the code snippet, and resume after the snippet is done. This is useful when running TUIs as otherwise those won't run given the terminal is taken by presenterm already.

Fixes #365

### DIFF
--- a/src/execute.rs
+++ b/src/execute.rs
@@ -54,45 +54,69 @@ impl SnippetExecutor {
         self.executors.contains_key(language)
     }
 
-    /// Execute a piece of code.
-    pub(crate) fn execute(&self, code: &Snippet) -> Result<ExecutionHandle, CodeExecuteError> {
-        if !code.attributes.execute && !code.attributes.execute_replace {
-            return Err(CodeExecuteError::NotExecutableCode);
-        }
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by orhun** [comment]
I was wondering if it would be also possible to see the output of the command on stdout while running something like:

````
```bash +exec +acquire_terminal
cargo run --manifest-path ratatui/Cargo.toml --example flex
```
````

So that we will be able to see the `cargo run` command running before it launches the TUI. Otherwise there is no indicator of whether if the build is ongoing or not.

PR #368, opened 2024-09-12
feat: add no_background property

### LINKED ISSUE #349: Best way to display multi-line blocks?
I've been playing around with `presenterm` for a coding presentation I have coming up, and I was wondering if there's a good way to render multi-line things like ASCII art without using images or potentially code blocks?

block_example

To be more specific, I want to use things like stylized ASCII text or QR codes in my presentation, but I can't find a satisfactory way to insert those in a way that I feel looks good. Putting them in code blocks like in the photo above works in terms of getting them to render in the presentation, but then they have the colored background; I can remove the background using themes/config, but I want the background for the actual code blocks, just not these "display" blocks, and I can't find a way to conditionally format anything in the documentation.

Apologies if this is a vague/strange question, I'm happy to clarify if that would help.

I can think of some ways to potentially implement this if it doesn't exist, but I'm not sure if any of them are idiomatic for the project, or if this idea is even within the scope of the project at all.

mfontanini:
This renames the `no_margin` attribute added in #363 to `no_background` and makes it do everything `no_margin` did but also prevents presenterm from adding a background to the snippet. This currently doesn't play well with `exec_replace` or `exec` meaning the snippet execution output still gets background. I will address this separately as there's other things I want to get to but I want to close this issue.

The following

~~~markdown
```text +no_background
.__    .__
|  |__ |__|   _____   ____   _____
|  |  \|  |  /     \ /  _ \ /     \
|   Y  \  | |  Y Y  (  <_> )  Y Y  \
|___|  /__| |__|_|  /\____/|__|_|  /
     \/           \/             \/
```
~~~

Currently renders like this after this change:

image

Fixes #349

### DIFF
--- a/src/processing/builder.rs
+++ b/src/processing/builder.rs
@@ -856,7 +856,7 @@ impl<'a> PresentationBuilder<'a> {
 
     fn code_style(&self, snippet: &Snippet) -> CodeBlockStyle {
         let mut style = self.theme.code.clone();
-        if snippet.attributes.no_margin {
+        if snippet.attributes.no_background {
             style.alignment = match style.alignment {
                 Some(Alignment::Center { .. }) => {
                     Some(Alignment::Center { minimum_size: 0, minimum_margin: Margin::default() })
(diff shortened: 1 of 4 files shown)


chris-t-jansen:
header_no_background

This worked perfectly for me, thank you so much! Exactly what I needed.

PR #387, opened 2024-10-30
feat: add tcl code highlighting

jtplaarj:
Added TCL code highlighting.


jtplaarj:
Sorry for the little mess, but I have added tcl and verilog in two different commits.

src/processing/code.rs
```diff
@@ -535,12 +537,14 @@ impl FromStr for SnippetLanguage {
...
             "svelte" => Svelte,
             "swift" => Swift,
+            "tcl" => Tcl,
             "terraform" => Terraform,
             "toml" => Toml,
             "typescript" | "ts" => TypeScript,
             "typst" => Typst,
             "xml" => Xml,
             "yaml" => Yaml,
+            "v" => Verilog,
```
mfontanini, line 547: This is the string you'll use in markdown to identify the snippet. I think this needs to be `verilog` no? At least github's markdown uses verilog. `v` is another language.

mfontanini:
Thanks!

Issue #112, opened 2023-12-29
Support for speaker notes


mfontanini:
It could be nice to support speaker notes. The idea would be to have a way to hook up another `<something>` (likely another instance of `presenterm`) which displays any speaker notes in the current slide, where notes would likely be some form of HTML comment. 

I don't personally use speaker notes, but I'm keeping this issue open so that those who want speaker note support state their expectations so we don't implement something blindly.

xstasi:
Hello there!

I make ample use of speaker notes and found the lack of support in presenterm a little bit distressing, so I cooked up a very simple "implementation" (emphasis on the quotes) using kitty.

Not claiming to be anywhere near a proper implementation, but perhaps others who stumble on this issue while looking around for speaker note support may find it useful.

Gist over [here](https://gist.github.com/xstasi/dd391f55e1c6b8795489919d25edec13)

Ta!

dmackdev:
I think speaker notes would be a useful addition to `presenterm`. I would definitely make use of them myself.

I experimented with a PoC of using 2 instances of `presenterm` to support speaker notes as follows:

- I added support for a speaker notes HTML comment: `<!-- speaker_note: Your speaker note here. -->`.
- I added a new optional CLI option: `--speaker-notes-mode`.
- `--speaker-notes-mode=publisher` is used to present slides. In this mode, speaker note comments are ignored, and every time the slide changes, an event is sent via inter-process communication (IPC).
- `--speaker-notes-mode=receiver` is used to view speaker notes. In this mode, only speaker note comments are rendered, and on each iteration of the render loop, a check is performed to see if there are any IPC events to change slide.
- If `--speaker-notes-mode` is not specified, no IPC structures are created, no events are sent/listened for, and speaker notes are ignored. 

Branch is here https://github.com/dmackdev/presenterm/tree/speaker-notes.

See a demo below. There is a slight layout issue on the second speaker note slide, but it's just a rough PoC at this stage:

https://github.com/user-attachments/assets/91614b29-14c6-4373-b2a5-08f384da008f

Keen to know your thoughts @mfontanini.

mfontanini:
Love the approach @dmackdev! Can you create a PR so we can go over the changes?

mikavilpas:
I like it! I think it would be very easy to embed the note viewer in a terminal inside my editor. It would fit my workflow really well 👍🏻

PR #389, opened 2024-11-06
feat: support speaker notes

dmackdev:
Closes https://github.com/mfontanini/presenterm/issues/112.

- Adds support for speaker notes via via inter-process communication (IPC) using [iceoryx2](https://github.com/eclipse-iceoryx/iceoryx2):
- Bumps CI `rust` version to 1.75.0 (for compatibility with `iceoryx2`).
- Adds new speaker notes example presentation.

## Speaker notes implementation:
- Speaker notes can be added to a presentation markdown file in the form of HTML comments: `<!-- speaker_note: Your speaker note here. -->`.
- `--speaker-notes-mode=publisher` CLI option is used to present slides. In this mode, speaker note comments are ignored and not rendered, and every time the slide changes, an IPC message is published.
- `--speaker-notes-mode=receiver` is used to view speaker notes. In this mode, only the title of each slide and speaker note comments are rendered. When retrieving the next command, we first check for an IPC message to change slides.
- If `--speaker-notes-mode` is not specified, no IPC structures are created, no messages are published/listened for, and speaker notes are ignored and not rendered.

## Demo:
https://github.com/user-attachments/assets/ee75452b-5ea5-4d02-a599-3d7e8d9f69c0

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ ansi-parser = "0.9"
 base64 = "0.22"
 bincode = "1.3"
 clap = { version = "4.4", features = ["derive", "string"] }
-comrak = { version = "0.29", default-features = false }
+comrak = { version = "0.31", default-features = false }
 crossterm = { version = "0.28", features = ["serde"] }
 directories = "5.0"
 hex = "0.4"
@@ -32,8 +32,7 @@ serde_with = "3.6"
 strum = { version = "0.26", features = ["derive"] }
 tempfile = "3.10"
 tl = "0.7"
-console = "0.15.8"
-thiserror = "1"
+thiserror = "2"
 unicode-width = "0.2"
 os_pipe = "1.1.5"
 libc = "0.2.155"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,7 @@ thiserror = "1"
 unicode-width = "0.2"
 os_pipe = "1.1.5"
 libc = "0.2.155"
-iceoryx2 = "0.4.1"
+iceoryx2 = "0.5.0"
 
 [dependencies.syntect]
 version = "5.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ serde_json = "1.0"
 serde_with = "3.6"
 strum = { version = "0.26", features = ["derive"] }
 tempfile = "3.10"
+tl = "0.7"
 console = "0.15.8"
 thiserror = "1"
 unicode-width = "0.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,6 +36,7 @@ thiserror = "1"
 unicode-width = "0.2"
 os_pipe = "1.1.5"
 libc = "0.2.155"
+iceoryx2 = "0.4.1"
 
 [dependencies.syntect]
 version = "5.2"
--- a/examples/code.md
+++ b/examples/code.md
@@ -6,13 +6,14 @@ theme:
       background: false
 ---
 
-# Code styling
+Code styling
+===
 
 This presentation shows how to:
 
-- Left-align code blocks.
-- Have code blocks without background.
-- Execute code snippets.
+* Left-align code blocks.
+* Have code blocks without background.
+* Execute code snippets.
 
 ```rust
 pub struct Greeter {
@@ -34,11 +35,10 @@ fn main() {
 }
 ```
 
-<!-- speaker_note: These are speaker notes on slide 1. -->
-
 <!-- end_slide -->
 
-# Column layouts
+Column layouts
+===
 
 The same code as the one before but split into two columns to split the API definition with its usage:
 
@@ -76,11 +76,10 @@ fn main() {
 }
 ```
 
-<!-- speaker_note: These are speaker notes on slide 2. -->
-
 <!-- end_slide -->
 
-# Snippet execution
+Snippet execution
+===
 
 Run code snippets from the presentation and display their output dynamically.
 
@@ -91,11 +90,10 @@ for i in range(0, 5):
     time.sleep(0.5)
 ```
 
-<!-- speaker_note: These are speaker notes on slide 3. -->
-
 <!-- end_slide -->
 
-# Snippet execution - `stderr`
+Snippet execution - `stderr`
+===
 
 Output from `stderr` will also be shown as output.
 
@@ -108,5 +106,3 @@ echo "This is a successful command again"
 sleep 0.5
 man # Missing argument
 ```
-
-<!-- speaker_note: These are speaker notes on slide 4. -->
--- a/examples/code.md
+++ b/examples/code.md
@@ -6,14 +6,13 @@ theme:
       background: false
 ---
 
-Code styling
-===
+# Code styling
 
 This presentation shows how to:
 
-* Left-align code blocks.
-* Have code blocks without background.
-* Execute code snippets.
+- Left-align code blocks.
+- Have code blocks without background.
+- Execute code snippets.
 
 ```rust
 pub struct Greeter {
@@ -35,10 +34,11 @@ fn main() {
 }
 ```
 
+<!-- speaker_note: These are speaker notes on slide 1. -->
+
 <!-- end_slide -->
 
-Column layouts
-===
+# Column layouts
 
 The same code as the one before but split into two columns to split the API definition with its usage:
 
@@ -76,10 +76,11 @@ fn main() {
 }
 ```
 
+<!-- speaker_note: These are speaker notes on slide 2. -->
+
 <!-- end_slide -->
 
-Snippet execution
-===
+# Snippet execution
 
 Run code snippets from the presentation and display their output dynamically.
 
@@ -90,10 +91,11 @@ for i in range(0, 5):
     time.sleep(0.5)
 ```
 
+<!-- speaker_note: These are speaker notes on slide 3. -->
+
 <!-- end_slide -->
 
-Snippet execution - `stderr`
-===
+# Snippet execution - `stderr`
 
 Output from `stderr` will also be shown as output.
 
@@ -106,3 +108,5 @@ echo "This is a successful command again"
 sleep 0.5
 man # Missing argument
 ```
+
+<!-- speaker_note: These are speaker notes on slide 4. -->
--- a/src/input/speaker_notes.rs
+++ b/src/input/speaker_notes.rs
@@ -2,4 +2,5 @@
 #[repr(C)]
 pub enum SpeakerNotesCommand {
     GoToSlide(u32),
+    Exit,
 }
--- a/src/input/speaker_notes.rs
+++ b/src/input/speaker_notes.rs
@@ -1,0 +1,5 @@
+#[derive(Debug)]
+#[repr(C)]
+pub enum SpeakerNotesCommand {
+    GoToSlide(u32),
+}
--- a/src/main.rs
+++ b/src/main.rs
@@ -348,6 +348,7 @@ fn run(mut cli: Cli) -> Result<(), Box<dyn std::error::Error>> {
 }
 
 fn main() {
+    iceoryx2::prelude::set_log_level(iceoryx2::prelude::LogLevel::Error);
     let cli = Cli::parse();
     if let Err(e) = run(cli) {
         eprintln!("{e}");
--- a/src/main.rs
+++ b/src/main.rs
@@ -215,7 +215,7 @@ fn create_speaker_notes_service_builder(
     presentation_path: &Path,
 ) -> Result<Builder<SpeakerNotesCommand, (), Service>, Box<dyn std::error::Error>> {
     let file_name = presentation_path.file_name().expect("failed to resolve presentation file name").to_string_lossy();
-    let service_name = format!("presenterm/{}", file_name).as_str().try_into()?;
+    let service_name = format!("presenterm/{file_name}").as_str().try_into()?;
     let service = NodeBuilder::new()
         .create::<Service>()?
         .service_builder(&service_name)
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,12 +1,15 @@
-use clap::{error::ErrorKind, CommandFactory, Parser};
+use clap::{error::ErrorKind, CommandFactory, Parser, ValueEnum};
 use comrak::Arena;
 use directories::ProjectDirs;
+use iceoryx2::{node::NodeBuilder, prelude::ServiceName, service::ipc};
 use presenterm::{
     CommandSource, Config, Exporter, GraphicsMode, HighlightThemeSet, ImagePrinter, ImageProtocol, ImageRegistry,
     MarkdownParser, PresentMode, PresentationBuilderOptions, PresentationTheme, PresentationThemeSet, Presenter,
-    PresenterOptions, Resources, SnippetExecutor, SpeakerNotesMode, Themes, ThemesDemo, ThirdPartyConfigs,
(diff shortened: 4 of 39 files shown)


dmackdev (commented):
This requires rust 1.75 or newer, but CI runs rust 1.74.
A rust-toolchain file and Cargo.toml entry would help make this more visible.

Cargo.toml
```diff
@@ -36,6 +36,7 @@ thiserror = "1"
 unicode-width = "0.2"
 os_pipe = "1.1.5"
 libc = "0.2.155"
+iceoryx2 = "0.4.1"
```
dmackdev, line 39: This requires rust 1.75 or newer, but CI runs rust 1.74.
A rust-toolchain file and Cargo.toml entry would help make this more visible.

dmackdev, line 39: This requires rust 1.75 or newer, but CI runs rust 1.74.
A rust-toolchain file and Cargo.toml entry would help make this more visible.

src/presenter.rs
```diff
@@ -104,6 +115,11 @@ impl<'a> Presenter<'a> {
             self.render(&mut drawer)?;
 
             loop {
+                if let Some(idx) = self.state.presentation().listen_for_speaker_note_evt() {
```
mfontanini, line 118: I think it would be interesting for this thing to be inside `input/source.rs` and act as a normal command, e.g. it could return `Command::GoToSlide` when such an event is found. This would make it more transparent to the presenter.

src/presentation.rs
```diff
@@ -292,6 +308,22 @@ impl PresentationState {
 
     fn set_current_slide_index(&self, value: usize) {
         self.inner.deref().borrow_mut().current_slide_index = value;
+        if let Some(SpeakerNoteChannel::Notifier(notifier)) = &self.inner.deref().borrow().channel {
```
mfontanini, line 311: Following the idea of moving this to presenter, you can always emit this event after a command is handled. That would help detach this logic from out of here.

mfontanini (commented):
A few comments mostly on whose responsibility this is but overall looks good!

examples/code.md
```diff
@@ -6,14 +6,13 @@ theme:
       background: false
 ---
 
-Code styling
-===
+# Code styling
```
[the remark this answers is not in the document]
dmackdev: Yes, I think my editor extension did this - I will revert this and add a separate example presentation as you suggested, with consistent syntax.

src/processing/builder.rs
```diff
@@ -418,6 +453,19 @@ impl<'a> PresentationBuilder<'a> {
             Ok(comment) => comment,
             Err(error) => return Err(BuildError::CommandParse { line: source_position.start.line + 1, error }),
         };
+
+        if let Some(SpeakerNotesMode::Receiver) = self.options.speaker_notes_mode {
+            match comment {
+                CommentCommand::SpeakerNote(note) => {
+                    self.push_text(note.into(), ElementType::Paragraph);
```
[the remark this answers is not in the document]
dmackdev: I initially did this because, for the speaker notes presentation, only these two comments are handled: speaker notes, and end slide. The other comments (e.g. layout, pauses, etc) need not be applied to the speaker notes presentation.

In general, I think my additions to this file need to be refactored.

Cargo.toml
dmackdev: Any thoughts on bumping the rust version @mfontanini? 
Otherwise a different version of `iceoryx2` (if one exists for rust 1.74) or different IPC crate will need to be used.

examples/code.md
```diff
@@ -35,10 +34,11 @@ fn main() {
 }
 ```
 
+<!-- speaker_note: These are speaker notes on slide 1. -->
```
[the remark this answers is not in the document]
dmackdev: Started a new speaker notes presentation.

src/presenter.rs
dmackdev: That would make sense - it _is_ a source of a command after all. 

So are you thinking:

- Remove the `SpeakerNoteChannel` enum,
- Store the `Option<iceoryx2::Listener>` directly in `CommandSource` if viewing speaker notes,
- Store the `Option<iceoryx2::Notifier>` directly in `Presenter` for the main presentation, if presenting with speaker notes.
?

Cargo.toml
dmackdev: Also need to investigate some warnings and proper configuration for usage of this crate.

src/processing/builder.rs
```diff
@@ -254,6 +260,19 @@ impl<'a> PresentationBuilder<'a> {
     }
 
     fn process_element(&mut self, element: MarkdownElement) -> Result<(), BuildError> {
+        if self.options.render_speaker_notes_only {
+            match element {
+                MarkdownElement::Comment { comment, source_position } => {
+                    self.process_comment(comment, source_position)?
+                }
+                MarkdownElement::SetexHeading { text } => self.push_slide_title(text),
```
dmackdev, line 268: I was thinking showing the slide title would be good for the speaker notes mode to help match it against the title of the current slide in the main presentation.

This also means we can support implicit slide ends in the speaker notes mode for free.

src/presenter.rs
mfontanini: Yes, exactly :ok_hand:

src/processing/builder.rs
mfontanini: I think it would be good to have a separate `process_element` for each mode so this one's not polluted with the logic from both. But I agree that having the heading makes sense :ok_hand:

src/presenter.rs
```diff
@@ -136,6 +158,10 @@ impl<'a> Presenter<'a> {
                     CommandSideEffect::None => (),
                 };
             }
+            if let Some(SpeakerNoteChannel::Notifier(notifier)) = self.speaker_notes_channel.as_mut() {
+                let current_slide_idx = self.state.presentation().current_slide_index();
+                notifier.notify_with_custom_event_id(EventId::new(current_slide_idx + 1)).unwrap();
```
mfontanini, line 163: Did you consider using the [interprocess](https://docs.rs/interprocess/latest/interprocess/index.html) crate instead? I need to read iceoryx2's docs more but this API feels a bit strange. You're communicating via event ids, which is enough for what we're doing but it feels a bit off. (I admit I haven't looked at the docs enough to understand what events mean here).

src/processing/builder.rs
```diff
@@ -253,7 +263,7 @@ impl<'a> PresentationBuilder<'a> {
         self.push_line_break();
     }
 
-    fn process_element(&mut self, element: MarkdownElement) -> Result<(), BuildError> {
+    fn process_element_for_presentation_mode(&mut self, element: MarkdownElement) -> Result<(), BuildError> {
```
dmackdev, line 266: Probably needs a better name so as to not be confused with `PresentMode`.

dmackdev:
Is there any reason why `nightly` toolchain is used for `cargo fmt`/`cargo clippy` in CI?

src/presenter.rs
dmackdev: Done.

src/processing/builder.rs
dmackdev: I have split this. `process_comment` could benefit from the same split, but there are extra processing on the raw comment string.

src/presenter.rs
dmackdev: No hard reason to keep using the `EventId` - `iceoryx2` also supports publish/subscribe messaging pattern with different payloads. The examples in the readme/examples dir of the repo show publishing a `usize`. We could easily use this pattern instead.

dmackdev: I had come across `interprocess`, but initially chose `iceoryx2` since it appeared to be more popular.
I have since tried the `interprocess` crate and had a quick go at using it in this branch but have not gotten that to work fully yet. Even running the local socket example from `interprocess` I noticed a few things: the "client" must strictly be started after the "server" otherwise it panics, and you have to manually remove the socket it creates between invocations, otherwise it panics.
So `iceoryx2` seems to be more polished in this regard, and I think has a nicer interface.

mfontanini: Fair, `interprocess` would also require dealing with all the clients which is pretty annoying.

mfontanini:
Looking at iceoryx2 I think you should really be using the [pub/sub mode](https://docs.rs/iceoryx2/latest/iceoryx2/#publish-subscribe). The event one you're using doesn't make sense to me, like I said it's communicating via event ids which only works here because you're passing in a number but it won't work if this needs to be changed in the future. 

What I'd suggest is the following:

* Use pubsub mode. This shouldn't require many changes to what you have.
* Send a specific `enum SpeakerNotesCommand` type using `serde_json` to serialize it to JSON. This type should have a single variant for now (`GoToSlide` or something) that contains a slide number. This will allow changing this in the future in case we want more types of commands to be sent. Actually typing this out now.... I wonder if this should just send commands directly and not have the listener parse the presentation at all. e.g. just send a `{"slide_title": "abc", "note": "this is where I talk about abcs"}` and let the other side display it. This would simplify the logic a lot since you wouldn't need to mix the listener into the `Presenter` type, but it would also require restructuring things quite a bit.
* Change the way the creation of each side of the pipe works:
  * The pipe/service name should contain some specific identifier so you can open more than one presentation at the same time. e.g. `presenterm test.md --publish-speaker-notes my-service`.
  * Creating the pipe on the publisher mode shouldn't use `open_or_create`, you want to explicitly create it and if it's open already it should fail (I expect this to be what that crate does). Otherwise having 2 presenterm instances running in publisher mode open will break or misbehave.
  * Creating the pipe on the listener mode should use `open` specifically, not `open_or_create`. I think what you mentioned for `interprocess` about the publisher needing to be up before the listener is fine, that's how it should be. I shouldn't be allowed to open the listener if no presentations are running in publish mode.

PS: the formatting needs nightly because most of the useful cargo formatting rules used here are nightly only.

dmackdev:
1. I have now updated the code to use the pubsub messaging pattern.
2. There are some restrictions on data types that can be sent in IPC messages with `iceoryx2` - see [here](https://github.com/eclipse-iceoryx/iceoryx2/tree/main/examples/rust/publish_subscribe_dynamic_data). Namely, "Data types like String or Vec will cause undefined behavior and may result in segmentation faults". So it would appear that heap allocated data (like serialised JSON strings) would require use of a fixed-sized container, which could be problematic. The current enum and its single variant that I have introduced `SpeakerNoteCommand::GoToSlide(u32)` works fine however, since it complies with the restrictions.
3. I have changed the pipe creation as suggested, and it works as expected - creating a duplicate publisher panics, and creating a listener before a publish also panics. For the service name, do you not think it would be a better user experience to infer this from the presentation file name/path, rather than requiring a user to specify a service name?

mfontanini:
I see, makes sense, I forgot this uses shared memory so there really isn't a reason to serialize to JSON anyway. I think if there's a need to use strings in the future they can always be self contained `[char; N]` types. It won't be _super_ flexible but flexible enough.

re: naming makes sense as well, not sure if there are restrictions on the length of the service name. Using the file name should be enough. e.g. `presenterm/demo.md` could be the name of the service if I run it like `presenterm examples/demo.md`.

dmackdev:
Updated the service name to incorporate the presentation file name.

There does not appear to be any length restrictions on the service name, only that it cannot be empty (see [here](https://github.com/eclipse-iceoryx/iceoryx2/blob/main/iceoryx2/src/service/service_name.rs#L35-L42)).

src/main.rs
```diff
@@ -270,9 +305,21 @@ fn run(mut cli: Cli) -> Result<(), Box<dyn std::error::Error>> {
             println!("{}", serde_json::to_string_pretty(&meta)?);
         }
     } else {
-        let commands = CommandSource::new(config.bindings.clone())?;
+        let speaker_notes_event_receiver = if let Some(SpeakerNotesMode::Receiver) = cli.speaker_notes_mode {
+            let receiver = create_speaker_notes_service_builder(&path)?.open()?.subscriber_builder().create()?;
```
mfontanini, line 309: This should fail with a nice looking error. I currently only see 

```
PublishSubscribeOpenError::DoesNotExist
````

When I run the receiver before the publisher. Maybe use a `Cli::command().error(...)` like above in this file with a meaningful error message like "no presenterm in publisher mode running" or something like that.

Same for the publisher failing when the channel is already open.

File: src/input/speaker_notes.rs [hunk 11]
```diff
@@ -0,0 +1,5 @@
+#[derive(Debug)]
+#[repr(C)]
+pub enum SpeakerNotesCommand {
```
mfontanini, line 3: Maybe there could be an `Exit` command sent when presenterm exits? That way they both would close simultaneously.

mfontanini (commented):
Just a few comments mostly on error handling. I noticed when you run it it prints

````
        2 [W] "Config::global_config()"
              | Default config file found but unable to read data, populate config with default values.
```

Do you see this too? Not sure where it's coming from.

dmackdev:
Yes I have been seeing that warning - I had mentioned it in the PR description.
It's since been fixed on the `iceoryx2` `main` branch, but would come in the next release.

I think it would be a blocker for this PR since it is not an ideal user experience - any suggestions?

src/input/speaker_notes.rs
dmackdev: Good idea.

src/main.rs
```diff
@@ -189,6 +211,19 @@ fn overflow_validation(mode: &PresentMode, config: &ValidateOverflows) -> bool {
     }
 }
 
+fn create_speaker_notes_service_builder(
+    presentation_path: &Path,
+) -> Result<Builder<SpeakerNotesCommand, (), Service>, Box<dyn std::error::Error>> {
+    let file_name = presentation_path.file_name().expect("failed to resolve presentation file name").to_string_lossy();
```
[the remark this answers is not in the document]
dmackdev: I've propagated a `Cli::command().error(...)` here instead.

src/main.rs
dmackdev: Agreed.
There are several different error enums - each with multiple variants - involved in this chain of method calls, which makes this tricky to isolate these particular variants.

I'll see what I can do.

mfontanini:
Sorry, I saw that when you created the PR but forgot when trying it out.

> I think it would be a blocker for this PR since it is not an ideal user experience - any suggestions?

Yyyyyeaaah I agree, I hope there's a release out soon. Do you have a link to the PR where this was removed? Depending on what they use we can potentially null out that output.

dmackdev:
That misleading warning was fixed in https://github.com/eclipse-iceoryx/iceoryx2/pull/438.

However we could also change the `iceoryx2` log level in `main` via `iceoryx2::prelude::set_log_level(LogLevel::Warn)`. I have tried this and it hides the warning.

BTW apologies for the sporadic updates - I have gotten quite busy recently, but should have more availability from next week.

src/input/speaker_notes.rs
dmackdev: Done.

src/main.rs
dmackdev: I have added a custom `IpcServiceError` enum and `From` implementations to convert the specific `iceoryx2` error enum variants for these two specific scenarios, in order to provide better error messages. This could be extended to cover more error cases if we wished. Let me know what you think.

dmackdev:
Hi @mfontanini, any thoughts on the latest changes? Namely:

- Improved error messages from `iceoryx2` service open/create errors,
- The `iceoryx2` warning is now gone thanks to the new published version.

mfontanini:
Hey @dmackdev! I've been away from home so I haven't had a chance to look at the PR again. I'll have a look and test this again when I'm back in a few days.

mfontanini:
I'm back! Thanks, this is great!

PR #393, opened 2024-11-13
Add R to executors

jonocarroll:
Thanks for this awesome project! If it's welcome, I added [R](https://www.r-project.org/) support to the executors. I've only tested this locally but it seems to work nicely.

Screen Recording 2024-11-13 at 1 17 32 PM (1)

[Rscript](https://search.r-project.org/R/refmans/utils/html/Rscript.html) is a command-line tool that's installed along with R, and should behave in a way that's compatible with your approach.


mfontanini:
Thanks!

PR #438, opened 2025-02-05
feat: add support for kitty's font size protocol

mfontanini:
This adds support for kitty's font size protocol (https://github.com/kovidgoyal/kitty/issues/8226) which allows printing characters that take up more than one cell. This feature will be available in kitty >= 0.40.0 and is currently only available in nightly builds.

This for now is only supported in a subset of the theme components, namely:

* The introduction slide's presentation title (`intro_slide.title.font_size`).
* The slide titles (`slide_title.font_size`).
* The headings (`headings.h*.font_size`).

Font sizes are only used if the terminal emulator supports it so this doesn't change anything for emulators other than kitty (or other implementors of the protocol). If you find this somehow breaks something, please create an issue.

For now all built in themes set `intro_slide.title.font_size=2` and `slide_title.font_size=2`. I think this looks a lot better this way but please do comment here if you don't think built in themes should come with these values set.

These are now the first 2 slides in the `demo.md` example:

https://github.com/user-attachments/assets/8d761d86-8855-498a-9766-5294cdae3b57

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,7 +2,8 @@ use crate::{
     code::snippet::SnippetLanguage,
     commands::keyboard::KeyBinding,
     terminal::{
-        GraphicsMode, emulator::TerminalEmulator, image::protocols::kitty::KittyMode, query::TerminalCapabilities,
+        GraphicsMode, capabilities::TerminalCapabilities, emulator::TerminalEmulator,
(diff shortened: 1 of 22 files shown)


mfontanini:
If someone uses this kitty >= 0.40.0 and feel like the default themes around this look odd please comment here!

Issue #447, opened 2025-02-12
Obsidian syntax compatibility question


0atman:
Hi @mfontanini, thanks again for making this.

My primary markdown environment is Obsidian, which works REALLY well with persenterm. While it uses GFM, there's a few syntax differences that cause friction with this workflow. The headliner is `![[wikilinks.jpg]]` support.

I wonder what your appetite is for adding features from Obsidian, such as wikilinks?

0atman:
Certainly I'd be keen to pitch in on this work, it seems only fair (and I'm a Rust developer). I have a small wishlist of syntax that I'd love to be supported in presenterm, which I'd be happy to make individual issues for if you'd like to track them.

mfontanini:
Markdown parsing is done via [comrak](https://docs.rs/comrak) so ideally, anything markdown specific would be added there and we'd use it here. They do seem to have wikilinks support and I just created a PR that adds that (#448). It weirdly doesn't use the syntax you mention here, but rather `[[url]]`. I've never used wikilinks before, does this work for you?

mfontanini:
>  I have a small wishlist of syntax that I'd love to be supported in presenterm

Feel free to create issues and contribute if you'd like to! I'm actively working on this project so any reasonable feature request I'd be happy to work on as well.

0atman:
> > I have a small wishlist of syntax that I'd love to be supported in presenterm
> 
> Feel free to create issues and contribute if you'd like to! I'm actively working on this project so any reasonable feature request I'd be happy to work on as well.

Fantastic, I'll try to be as helpful as I can

PR #448, opened 2025-02-12
feat: add support for wikilinks

mfontanini:
This adds wikilinks support via the `[[url]]` syntax.

Relates to #447

### DIFF
--- a/src/markdown/parse.rs
+++ b/src/markdown/parse.rs
@@ -32,6 +32,7 @@ impl Default for ParserOptions {
         options.extension.strikethrough = true;
         options.extension.multiline_block_quotes = true;
         options.extension.alerts = true;
+        options.extension.wikilinks_title_before_pipe = true;
         Self(options)
     }
 }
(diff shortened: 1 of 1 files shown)


0atman:
Works for normal links, doesn't work for embedded images (which perhaps was not the scope of this PR).

```md
Test Pattern
===

[[presentest#Layout example]]

![[testpattern.jpg]]
```

Renders like this:

Screenshot-1739364844

0atman:
Whereas I was hoping for the image to be embedded like this:

Screenshot-1739364799

0atman:
Of course, you could argue that wikilink embedding (`![[]]`) is out of specification (it's not even in GFM), but I contend that due to the popularity of tools like Obsidian, it would be an EXTREMELY handy option, nonetheless!

mfontanini:
Hmm it would be ideal if comrak supported this. Otherwise on our end it's a bit hacky to do this. I'll see if this works but could you create an issue in their repo so it's supported natively?

mfontanini:
This is too annoying to do here unfortunately. The `![[link]]` syntax is not even recognized as a `!` followed by a wikilink by comrak. If that were the case I could check for that `!` and transform it into an image but with this we'd have to manually re-parse the text to find these `![[link]]` structures. I've seen good movement on comrak's repo so it's worth a try!

0atman:
> we can tackle images separately

Absolutely, seems non-trivial

PR #452, opened 2025-02-14
chore: bump sixel-rs version

Xeonacid:
Includes https://github.com/orhun/sixel-rs/pull/6 to fix build on aarch64 and riscv64.

If possible, please make a new release so downstream maintainers like Arch Linux RISC-V won't need to maintain this patch for an extended period. :P


mfontanini:
Thanks! I'm hoping to have a release out in a couple of weeks!

mfontanini:
Another bug came up and I cut a release including that one and this fix https://github.com/mfontanini/presenterm/releases/tag/v0.10.1

PR #488, opened 2025-03-10
feat: output the name of the theme selected with --list-themes to stderr

benoitlx:
This PR is not intended to be merged, as I might be the only one needing this feature, but I'm posting it if someone wants to take a look.

I find it useful to quickly be able to select a theme before a presentation. However, the `--list-themes` option does not allow for automating this process because it's not outputting anything.

This is why I came up with the idea to output the selected theme (the last one we see before quitting `presenterm --list-theme`) to stderr (not to stdout because of the dynamic nature of the `presenterm --list-theme` command, which would interfere if we needed to pipe the command's output).

Below is a `justfile` that I use to fulfill my needs:
```just
theme:
  presenterm --list-themes 2> theme

present:
  presenterm --theme $(cat theme) examples/demo.md
```

### DIFF
--- a/src/presentation/builder/mod.rs
+++ b/src/presentation/builder/mod.rs
@@ -720,7 +720,7 @@ impl<'a> PresentationBuilder<'a> {
 
         let incremental_lists = self.slide_state.incremental_lists.unwrap_or(self.options.incremental_lists);
         let iter = ListIterator::new(list, start_index);
-        if self.options.pause_before_incremental_lists {
+        if incremental_lists && self.options.pause_before_incremental_lists {
             self.push_pause();
         }
         for (index, item) in iter.enumerate() {
@@ -729,7 +729,7 @@ impl<'a> PresentationBuilder<'a> {
             }
             self.push_list_item(item.index, item.item)?;
         }
-        if self.options.pause_after_incremental_lists {
+        if incremental_lists && self.options.pause_after_incremental_lists {
             self.push_pause();
         }
         Ok(())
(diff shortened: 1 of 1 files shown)


mfontanini:
Hi @benoitlx, thanks for bringing this up! I just created #489 that achieves the same thing but instead adds a new `--current-theme` that prints the theme in use to stdout and exits. I think this is a bit cleaner so we don't overload the meaning of `--list-themes`. Does that work for you?

benoitlx:
Hi, thanks for taking a look at it! I'm not sure how I can use the feature of #489 for my use case. 

What I want is a theme selector that pops up before launching my actual presentation. I thought to achieve that by getting the name of the theme via my "patched" `--list-themes` and then launching my presentation with `--theme {{ name }}`. But, yes, I strongly agree that my initial solution is not clean at all since it overloads the meaning of `--list-themes`.

A better solution for my use case could be the following: 
- launch `run.demo()?` when a special argument (e.g. `choose`) is specified to the `--theme` option
  - or maybe with a different option like `--select-theme`
- let the user interact with the demo and capture the theme name when the user input `q` or `<C-C>`
- run the user's presentation with the selected theme

I hope I clarified my use case, and if you think my new idea is sensible I'll be glad to work on it (I think it'll be a good entry point to open source 🙂)!

mfontanini:
Sorry, I misunderstood what you meant and I get it now. I merged the other PR but I agree it doesn't help you solve your problem.

I'm not sure what's the best solution here... I'm thinking the `--list-themes` route may not be that bad...

I'm curious about your use case: do you change themes on the fly the moment you want to present, every time you present? In general all presentations I've done I choose the theme ahead of time and the presentation gets made with that theme in mind already. e.g. images are light/dark themed to fit the theme where appropriate.

benoitlx:
Apologies for the delay.

I didn't intend to close this PR; I was trying to sync my fork and accidentally deleted my commit (maybe I should revert this ?). 🙃

To provide some context on my use case, I often need to change the presentation theme on the fly. Currently, I have to exit the presentation and relaunch it with the appropriate `--theme` option. As you mentioned, this can lead to some images not fitting the new theme. To mitigate this, I export my images in both light and dark versions, and when an image doesn't match the theme, I manually edit the markdown file.

I have a couple of ideas that could enhance this process:
- Introduce a key binding and a corresponding entry in the presenter menu labeled "Theme" or whatever.
- When the user activates the key binding, it would trigger `run.demo()`, allowing them to select a theme.
- For images, we could implement a markdown template variable, such as `{{ color }}`, which would dynamically switch between dark and light.

I believe it would be beneficial to discuss these ideas further in a new issue, as they diverge from the original scope of this comment.

PR #495, opened 2025-03-13
Correctly include layout pic

Tuxified:
Seems like the example layout picture wasn't referred correctly and hence doesn't show up when visiting https://mfontanini.github.io/presenterm/features/layout.html . 

This change should probably fix it.


mfontanini:
Thanks!

PR #533, opened 2025-04-02
fix: fixing the external_snippet test error

Medovi:
The absolute path to the temporary directories has been replaced with the path specified in the environment variables.


mfontanini:
Thanks!

PR #559, opened 2025-04-21
Update highlighting.md (add julia)

younes-io:


younes-io:
thank you for this crate @mfontanini ! I was wondering which lib do you use for syntax highlighting & if it actually supports Julia

mfontanini:
Thanks! This is fixed in master now.

PR #560, opened 2025-04-22
feat: add `collapse_horizontal` slide transition

mfontanini:
This adds a new `collapse_horizontal` slide transition that collapses the current slide into the center of the screen.

https://github.com/user-attachments/assets/a398c286-ae3e-4c99-81aa-26aba498726d

### DIFF
--- a/config-file-schema.json
+++ b/config-file-schema.json
@@ -586,6 +586,22 @@
             }
           },
           "additionalProperties": false
+        },
+        {
+          "description": "Collapse the current slide into the center of the screen.",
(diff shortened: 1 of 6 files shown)


0atman:
Classic! Feels like some classic Star Wars transitions :smile:  
Btw, to get these bleeding edge features, do we just `cargo install --git`?

mfontanini:
Yes, that should work. There's also a nightly release you can find in the releases pages. I'm hoping to create a new release in the next few days.

PR #566, opened 2025-04-24
feat: allow exporting to html

JustSimplyKyle:


src/export/output.rs
```diff
@@ -122,17 +124,29 @@ impl ContentManager {
     }
 }
 
-pub(crate) struct PdfRender {
+pub(crate) enum OutputType {
```
mfontanini, line 127: `OutputFormat` maybe?

src/main.rs
```diff
@@ -68,15 +68,19 @@ struct Cli {
     path: Option<PathBuf>,
 
     /// Export the presentation as a PDF rather than displaying it.
-    #[clap(short, long)]
+    #[clap(short, long, group = "export")]
     export_pdf: bool,
 
+    /// Export the presentation as a HTML rather than displaying it.
+    #[clap(short, long, group = "export")]
```
mfontanini, line 75: This is going to fail because both options start with an `e`. You can drop the `short` option here for now.

src/export/script.js
```diff
@@ -0,0 +1,42 @@
...
+      line.classList.add('hidden');
+    });
+
+    allLines[currentPageIndex].classList.remove('hidden');
+  }
+
+
+  function scaler() {
+    var w = document.documentElement.clientWidth;
+    let scaledAmount= w/829;
```
mfontanini, line 18: What is this magical number?

mfontanini (commented):
This is awesome work!! I wanted to get to HTML exports soon but I didn't realize it could be built so well on the existing export code. I thought this would require an HTML page not tied to pixels but it seems you made it work. Thanks!!

src/export/script.js
JustSimplyKyle: it's the original width of the presentation, I left it in there for testing
I forgot to actually implement it :shrug:

mfontanini:
PDFs are still cut with the latest commit. I think the problem is in embedding the CSS in the HTML file. I initially tried that but it seemed like weasyprint wouldn't do the right thing unless I put it in a separate file and included it via the `-s` parameter, that's why that was that way.

mfontanini:
Sorry, still broken :(. Looks like the issue is not the `-s` (or not _just_ that); I've traced this back to [this commit](https://github.com/mfontanini/presenterm/pull/566/commits/c3fb212f9043e042dcd254cb03febdb664255b1d). The previous commit works, this one doesn't. I think the styles will probably need to be generated dynamically so only the HTML version contains those `with` and other attributes.

Not sure what environment you're using; I'm surprised it's not breaking at all for you. Are you using macOS?

JustSimplyKyle:
I'm using arch linux.

After re-checking your image, I think it may be linked to the `transform-origin` being set to top center
So I reworked it so that it's set to top left

If this doesn't work, I'll try separating the two then

mfontanini:
Sorry I couldn't have a look yesterday, this is still broken :(

JustSimplyKyle:
it seems to be an upstream bug
when I downgrade my weasyprint to 62.3, the bug gets reproduces
but when on 63.1, the bug resolves itself

anyway, with the latest commit, both 62.3 and above works without issue

mfontanini:
Thanks so much, this works great!

Issue #604, opened 2025-05-22
Can footnotes be superscript?


0atman:
Firstly, I'm delighted footnotes are handled at ALL! Thank you, not everyone cares about them.

## Problem

However, the footnote indicator is displayed as-is:

```markdown
Footnotes ideally would be superscript[^1]
```

Renders as:

![Image](https://github.com/user-attachments/assets/0aa0b4b4-a1b7-4fe1-bcad-ee7a27d5d4bb)

Whereas typically they would be rendered as superscript, like here in my static site generator:

![Image](https://github.com/user-attachments/assets/549c444b-94eb-4803-bbee-071aab018a8a)

## Proposed Solution

Unicode superscript numerals:

```markdown
Perhaps unicode superscript numerals could be used?¹
```

Renders in presenterm as:

![Image](https://github.com/user-attachments/assets/6aab8ff4-1d3e-4b2a-ae75-080f783e8d89)

Perhaps there's a better way to do it, but this is pretty simple, at least!

Thanks as ever for the great work here 😃

mfontanini:
Created #606 as an alternative solution for this: supporting superscript syntax (e.g. `f(x) = x^2^ + y^3^`) via the kitty font size protocol (using half blocks). My worry with supporting footnotes is that I imagine I'd need to decide where to place them. e.g. Just letting you _use_ footnote syntax feels like doing too little from presenterm's side but it really should place the footnote definitions at the bottom of the slide. However, I don't really want presenterm to have to figure out where exactly to place them.

By letting you use superscripts you can now achieve the same like:

```
Footnotes ideally would be superscript ^1^

...
...
...

^1^ https://github.com/mfontanini/presenterm/issues/604
```

Can you lmk what you think? It is straightforward after this superscript change to allow footnotes, I'm just worried about what to do with them.

0atman:
Footnotes are a semantic thing in markdown, so while we can fake it using the new superscript notation (thanks btw for that!) it would be great to get them to render using the actual markdown syntax.

Regarding where to place them - I have good news: At least in my use case, I would prefer to leave footnotes where they are - don't hoist them to the slide they are used.

Example:

```markdown
First slide
===

What a great slide![^1]

Second slide
===

Another great slide![^2]

Third slide
===

Nothing here.

Footnotes
===

[^1]: Can confirm
[^2]: Masterfull
```

[...]

PR #606, opened 2025-05-22
feat: support superscript via kitty font size protocol

mfontanini:
This adds support for superscripts via `x^42^` syntax. This requires the terminal to support half sizes (`n=X:d=Y`) via the kitty font size protocol. This works _okay_ in PDF/HTML exports, except when using superscripts in block quotes. In that case it looks a bit messed up. It's possible there's other places where it also looks bad (now typing this, alerts are also affected).

This is using half blocks but also still making them take up an entire cell. There's a way around this but requires changing a few things so I'd rather have something functional for now that can be iterated on.

Relates to #604

### DIFF
--- a/src/export/html.rs
+++ b/src/export/html.rs
@@ -8,8 +8,9 @@ pub(crate) enum HtmlText {
 
 impl HtmlText {
     pub(crate) fn new(text: &str, style: &TextStyle, font_size: FontSize) -> Self {
+        let mut text = text.to_string();
         if style == &TextStyle::default() {
-            return Self::Plain(text.to_string());
+            return Self::Plain(text);
         }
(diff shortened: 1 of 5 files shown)


0atman:
Nice to have this feature! however, afaik, the most standard way to do superscript in markdown is simply with `<sup>1</sup>` html - I wonder if a special case could be added for this tag, like the span styles for colour?

mfontanini:
Hey @0atman, I somehow missed your comment and only found it now while preparing for another release. I added support for `sup` in #665

0atman:
@mfontanini oh amazing! small feature, perhaps, but a huge bonus - thank you!

PR #609, opened 2025-05-24
feat: Allow 3 Digit Hex Colours and Some Clippy Auto-fixes

peterc-s:
Usually, CSS allows you to use 3 digit hex codes for colours (i.e. #ABC = #AABBCC). I've added support for this and also applied some Clippy auto-fixes.

### DIFF
--- a/src/theme/raw.rs
+++ b/src/theme/raw.rs
@@ -920,7 +920,7 @@ pub(crate) enum ParseColorError {
     #[error("invalid hex color: {0}")]
     Hex(#[from] FromHexError),
 
-    #[error("Hex color should only be 3 or 6 long, got hex string of length {0}.")]
+    #[error("hex color should only be 3 or 6 long, got hex string of length {0}")]
     InvalidHexLength(usize),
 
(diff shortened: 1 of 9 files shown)


src/theme/raw.rs
```diff
@@ -915,6 +920,9 @@ pub(crate) enum ParseColorError {
     #[error("invalid hex color: {0}")]
     Hex(#[from] FromHexError),
 
+    #[error("Hex color should only be 3 or 6 long, got hex string of length {0}.")]
```
mfontanini, line 923: Can you use no capital on Hex and no period at the end like the rest of the errors?

peterc-s:
> Thanks! I'd suggest not fixing the lints because that'll require bumping the rust version used in a couple of places. I've been meaning to do that, I'm seeing those clippy complaints locally as well

Would you like me to revert them then? Didn't realise it would require any version changes.

mfontanini:
Thanks! I'll fix the clippy lints and bump the rust version now

PR #616, opened 2025-05-26
feat: add support for footnotes

### LINKED ISSUE #604: Can footnotes be superscript?

mfontanini:
This adds support for markdown footnotes. Footnotes are written where they show up, as is. Example:

```
foo![^1] and bar[^2]

[^1]: hi
[^2]: bye
```

image

Closes #604 

PS: this requires the kitty font size protocol. In another PR I'll add something to use unicode half characters as fallback where possible.

### DIFF
--- a/src/markdown/elements.rs
+++ b/src/markdown/elements.rs
@@ -65,6 +65,9 @@ pub(crate) enum MarkdownElement {
         /// The content lines in this alert.
         lines: Vec<Line<RawColor>>,
     },
+
+    /// A footnote definition.
+    Footnote(Line<RawColor>),
 }
 
 #[derive(Clone, Copy, Debug, Default)]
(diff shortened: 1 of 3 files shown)


0atman:
YES! This is fantastic, thank you so much!

Screenshot-1748329603

Screenshot-1748329615

Issue #639, opened 2025-06-17
Validate Lint warnings


0atman:
Thanks again @mfontanini for `--validate-snippets`, it's REALLY sped up my workflow! 

**But I wonder if I can configure it to fail on warnings as well as errors?**

Here's an example I was just working on:
![Image](https://github.com/user-attachments/assets/02a01f60-5ab3-4d18-93b5-83eec8432fb6)

I'm running that presentation with `--validate-snippets`, and it correctly validated, no compiler errors, but when I execute the snippet with `c-e`, the warning shows.

EVEN BETTER would be for me to be able to tell `presenterm` to use `cargo clippy`, rather than `cargo build`, to keep me from making embarrassing mistakes 😆 

(I wonder if an in-line `cargo.toml` might help with this, to save re-configuring lints on each code block?)

Thank you!

mfontanini:
I just created #640 which adds a new `rust-script-pedantic` executor (e.g. `+exec:rust-script-pedantic`) which causes warnings to be treated as errors.

As for clippy, I don't think there's an easy way to make it work with rust-script. Probably it's doable but making it work cross platform will be a pain. Hopefully clippy can be integrated better with scripts once cargo script (https://github.com/rust-lang/cargo/issues/12207) is stabilized.

0atman:
Ah, yes, of course, clippy couldn't handle the deps. Yeah, cargo script maybe!

PR #640, opened 2025-06-17
feat: add `rust-script-pedantic` executor for rust

mfontanini:
This adds a new `rust-script-pedantic` alternative executor for rust. This is the same as `rust-script` but it passes `--deny warnings` to `rust-script` so any warnings are considered errors. This is useful when using `--validate-snippets` so you know there's no warnings nor errors in the snippets.

Relates to #639


0atman:
(At this point, I update presenterm from git nearly every day to get COOL STUFF! thank you!)

PR #645, opened 2025-06-19
feat: allow validating snippets without executing them in presentation

### LINKED ISSUE #643: Hide exec [not started] divider
When combined with `--validate-snippets`, my workflow is that I only execute the snippets while testing, not for final presentation:

Image

Which means I have to mentally reformat my slides to ignore the 3 lines after the code block. (I plan to remove the `+exec` from each one before presenting)

Is there a way to have my cake (validated snippets) and eat it (have the final version of the slide without `[not started]`), too?

Thanks very much @mfontanini again, snippet validation has REALLY simplified my work! <3

mfontanini:
This adds a `+validate` (also `+validate:<executor>`) that allows indicating that the snippet should be validated. This causes `--validate-snippets` to validate any snippet that's either `+exec` or `+exec_replace` or `+validate`.

This renames the `+validate` created in #637 that allows specifying whether the snippet should succeed or fail into `+expect`. e.g. `+expect:failure` means "this snippet should fail when being validated".

Closes #643

### DIFF
--- a/src/code/snippet.rs
+++ b/src/code/snippet.rs
@@ -246,6 +246,11 @@ impl SnippetParser {
                     attributes.representation = SnippetRepr::ExecReplace;
                     attributes.execution = SnippetExec::Exec(spec);
                 }
+                Validate(spec) => {
+                    if matches!(attributes.execution, SnippetExec::None) {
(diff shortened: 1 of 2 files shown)


0atman:
Perfection!

PR #662, opened 2025-07-01
Added uv as an executor for python code

JanNeuendorf:
With this you can use [uv](https://github.com/astral-sh/uvl) to execute python code by changing the language name to "uv". 
This calls `uv run --script -q (file)`. You can use the syntax for script dependencies:
```python
/// # /// script
/// # requires-python = ">=3.12"
/// # dependencies = [
/// #     "numpy",
/// # ]
/// # ///
import numpy as np
print("Hello from ts.py!")
```
and then use the triple slashes to hide it in the presentation.

### DIFF
--- a/executors.yaml
+++ b/executors.yaml
@@ -79,6 +79,11 @@ python:
   commands:
     - ["python", "-u", "$pwd/snippet.py"]
   hidden_line_prefix: "/// "
+  alternative:
+    uv:
+      filename: "snippet.py"
+      commands:
+        - ["uv", "run", "--script", "-q", "$pwd/snippet.py"]
 r:
   filename: snippet.R
   commands:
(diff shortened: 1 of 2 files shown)


executors.yaml
```diff
@@ -119,6 +119,11 @@ sh:
   commands:
     - ["sh", "$pwd/script.sh"]
   hidden_line_prefix: "/// "
+uv:
```
mfontanini, line 122: Can you add this instead as an `alternative` executor for `python`? See https://github.com/mfontanini/presenterm/blob/master/executors.yaml#L103-L104. By doing this you can still use `python` as the language in your snippet (so you hopefully get highlighting in your editor) but you just have to specify how the program is ran:

~~~markdown
```python +exec:uv
print("hi")
```
~~~

executors.yaml
```diff
@@ -119,6 +119,11 @@ sh:
...
+  filename: snippet.py
+  commands:
+    - ["uv", "run","--script","-q", "$pwd/snippet.py"]
```
mfontanini, line 125: nit: can you add proper spacing in between array elements here?

Issue #213, opened 2024-02-17
Remote images no longer loading


sachinsenal0x64:
![image](https://github.com/mfontanini/presenterm/assets/127573781/f3dc7a4c-de99-4072-a07c-77b94100706d)

image-protocol = kitty-remote

mfontanini:
That never worked and I don't think it's a good idea to do so. That makes presentations irreproducible and also can cause your live presentation to fail to load because of e.g. a shitty Internet connection.

sachinsenal0x64:
Ok, thanks for the quick reply.

PR #664, opened 2025-07-05
doc: clarify why no remote URLs work with images

ryuheechul:
addresses the need of having to search or create issues like https://github.com/mfontanini/presenterm/issues/213#issuecomment-1950342423


mfontanini:
Thanks!

mfontanini:
Added an explicit error when using http(s) URLs in image tags in #666

PR #680, opened 2025-07-11
feat: add support for multiline subtitle

barr-israel:
Using slightly modified code from #679 to also allow multiline subtitles.

One thing I didn't have a solution for was the build error in line 94, which will report a PresentationTitle error when there is an error parsing the subtitle, because there is no PresentationSubtitle error, and there is no clean way to separate the cases.  
I would argue that this is just a parsing error and the location it happened at should be passed inside it.


mfontanini:
Nice, thanks!

PR #682, opened 2025-07-11
feat: add support for multiline slide titles

barr-israel:
`SetexHeading` is now a vector of lines, with an implicit line break between them, inserted by `push_slide_title`.  
I decided to only put the contents of the first exHeading line as the slide name(for the slides modal), as there is no solution that will fit all formatting(without any additional insertions, the words will be squished together, with any added characters, it will look weird if the writer already added one)

### DIFF
--- a/src/demo.rs
+++ b/src/demo.rs
@@ -124,7 +124,7 @@ impl ThemesDemo {
             &parser,
             options,
         )?;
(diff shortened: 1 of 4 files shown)


mfontanini:
Nice, thanks!

PR #719, opened 2025-09-02
fix: don't add extra heading lines depending on font size

### LINKED ISSUE #714: `font_size` should change font of the line that it is declared
`font_size` only changes the font size on the line after, not the line itself, which I found surprising.

### What I Did

I'm trying to make a large H2, a smaller H3 and then regular body text underneath:

```markdown
<!-- font_size: 3 -->
## TITLE
<!-- font_size: 2 -->
### Subtitle
<!-- font_size: 1 -->
> TODO
```

### What I Expected

I expected the headings to be bunched up together, like they are if you set these sizes in the theme file:

```yaml
  h2:
    font_size: 3
  h3:
    font_size: 2 
```

Image.

### What Happened

Image

The font size change only takes place on the line AFTER, meaning it's never possible to affect the newline after the heading.

I get why this is *technically* reasonable, the `\n` occurred before the comment, but I believe it's suprising from a user's point of view? 

The current workaround is to go back to setting sizes in my theme file - but I'd like to separate size from semantics, so my document is semantically correct, but with a presentation where I can hand-craft each element's individual size.

#Proposed Fix

I propose these comments should apply to the line they are on, in addition to subsequent lines. Should someone want the existing behaviour, `<!-- new_line -->` is available to provide that.

Thanks as always! 😁

mfontanini:
This changes the behavior when an explicit `font_size` is set and a heading other than a slide title is used, so that we only add a single extra line after the heading, rather than one line per font size (e.g. 3 lines for font size 3). This was a deliberate choice initially but I'm not exactly sure why. If this breaks someone's flow please comment here!

Closes #714

### DIFF
--- a/src/presentation/builder/heading.rs
+++ b/src/presentation/builder/heading.rs
@@ -58,7 +58,7 @@ impl PresentationBuilder<'_, '_> {
         text.apply_style(&style.style);
 
         self.push_text(text, element_type);
-        self.push_line_breaks(self.slide_font_size() as usize);
+        self.push_line_break();
         Ok(())
     }
 }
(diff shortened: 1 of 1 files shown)


0atman:
YES DUDE! :tada: 

```markdown
<!-- font_size: 6 -->
# `OBSIDIAN`
<!-- font_size: 4 -->
# FOR STUDENTS
```
Screenshot-1756898888


Thank you so much for fixing this for me, I'm literally about to record, so this has saved so much faffing around with having different headings for the combined presenterm/web article!

The only negative is that I've reset my upgrade scripts to grab presenterm git latest - meaning more compiles for me :laughing:

PR #722, opened 2025-09-05
docs: sync supported terminals

gcavelier:
Just a small patch to have the same lists of supported terminals in `docs/src/features/images.md`, `examples/demo.md` and `README.md`


mfontanini:
Thanks!

Issue #725, opened 2025-09-10
Tokyonight Night Theme


cloudlena:
It's great to have the `tokyonight-storm` theme. However, it would be even better to also have the other variants of TokyoNight. Especially the default one (TokyoNight Night): https://github.com/folke/tokyonight.nvim

I would love to add the following ones:
- `tokyonight-night`
- `tokyonight-day`
- `tokyonight-moon`

If y'all agree, I'm happy to provide a PR.

PR #751, opened 2025-10-03
Add more Tokyo Night theme variants

cloudlena:
- Tokyo Night Moon
- Tokyo Night Night
- Tokyo Night Day

Fixes https://github.com/mfontanini/presenterm/issues/725

### DIFF
--- a/themes/tokyonight-day.yaml
+++ b/themes/tokyonight-day.yaml
@@ -1,0 +1,150 @@
+---
+default:
+  margin:
+    percent: 8
+  colors:
+    foreground: "3760bf"
+    background: "e1e2e7"
+
+slide_title:
+  alignment: center
+  padding_bottom: 1
+  padding_top: 1
+  colors:
+    foreground: "8c6c3e"
+  bold: true
+  font_size: 2
+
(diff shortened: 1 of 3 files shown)


themes/tokyonight-day.yaml
```diff
@@ -0,0 +1,150 @@
...
+      foreground: "9854f1"
+  h6:
+    prefix: "░░░░░░░"
+    colors:
+      foreground: "9854f1"
+
+block_quote:
+  prefix: "▍ "
+  colors:
+    foreground: "3760bf"
```
mfontanini, line 108: This is a bit illegible. You can use `presenterm --list-themes` to see how some basic slide would look like. The `alert` below probably needs the same set of colors.

Image

mfontanini (commented):
Thanks for setting this up! Maybe it's too late but you could have used a [palette like in this theme](https://github.com/mfontanini/presenterm/blob/master/themes/dark.yaml#L152-L165) and then reference colors via `palette:<color>`. This way you would only define colors once. It's not a huge deal, it would have just made your life easier :).

PR #773, opened 2025-11-02
feat: add support for user comments in presentation rendering

Cyberistic:
I've found the function `should_ignore_comment` and have been relying on it to add user comments inside presentations by abusing the `<!-- vim: comment here -->` since I don't use vim for this anyways.

This isn't clean and a dedicated way to comment should be added IMO. Also, there is an issue with the hacky vim solution above, it adds a newline where the comment should be ignored; fixed by using `self.slide_state.ignore_element_line_break = true` on ignored comments.

Current comments:
image

PR with ignore line breaks:
image

### DIFF
--- a/src/presentation/builder/comment.rs
+++ b/src/presentation/builder/comment.rs
@@ -152,18 +152,7 @@ impl PresentationBuilder<'_, '_> {
         } else {
             // Ignore vim-like code folding tags
             let comment = comment.trim();
-            if comment == "{{{" || comment == "}}}" {
-                return true;
-            }
-
-            // Ignore user comments with // prefix
-            // e.g., <!-- // This is a comment -->
-            let trimmed_comment = comment.trim_start_matches(&self.options.command_prefix).trim();
-            if trimmed_comment.starts_with("//") {
-                return true;
-            }
-
-            false
+            comment == "{{{" || comment == "}}}" || comment.starts_with("//")
         }
     }
 


mfontanini (commented):
Sorry, I've been out for a while and this slipped through. A couple of comments but looks good overall!

src/presentation/builder/comment.rs
```diff
@@ -149,7 +152,18 @@ impl PresentationBuilder<'_, '_> {
...
             // Ignore vim-like code folding tags
             let comment = comment.trim();
-            comment == "{{{" || comment == "}}}"
+            if comment == "{{{" || comment == "}}}" {
+                return true;
+            }
+
+            // Ignore user comments with // prefix
+            // e.g., <!-- // This is a comment -->
+            let trimmed_comment = comment.trim_start_matches(&self.options.command_prefix).trim();
```
mfontanini: Probably should have noticed this before but why the `trim_start_matches(command_prefix)`? This means you're trying to trim a prefix (say `cmd: ` although an empty string by default) and then trying to see if what you end up with starts with `//`. This means this code would ignore a command like `<!-- cmd: // foo -->` and that seems wrong to me.

e.g. test with this presentation:

```markdown
---
options:
  command_prefix: "cmd:"
---

<!-- cmd://hi -->
```

To me this line shouldn't exist and the line below should just check `if comment.starts_with("//")` since this is already trimmed a few lines above.

PR #778, opened 2025-11-17
feat: add dynamic theme option based on terminal color

JOTSR:
# Motivation
It is common to use light/dark theme switch at os level (for accessibility, eye care, ...) but it often needs to update theme config for TUI apps. 
Including theme detection is a real comfort especially for theme like `terminal-dark`/`terminal-light` that have no background colors.

# Description

I update config to support either
```yml
theme: my-theme
```
or
```yml
theme:
    dark: my-dark-theme
    light: my-light-theme
```
Theme detection use [`termbg`](https://crates.io/crates/termbg) crate

# Discussion

- corresponding cli flags can be added, I omit them as I don't see the direct usage
- a way to force specific theme (eg: `$THEME_MODE`, or `--theme-mode`), but it is already covered by `--theme`
- I can implement `Default` trait instead of using const for timeout, in this case what are the defaults
- I can change default theme from `dark` to dynamic theme (eg: `terminal-light`/`terminal-dark`)
- does the naming/code convention are ok
- I can change default timeout value, `100ms` seems large and is not perceptible (`termbg` latency is under 1ms on my terminals)
- I can handle terminal detection errors differently or change detection failed message (any suggestion)
- theme is only detected at launch, maybe I can add dynamic theme switch during presentation if it faisable (need to investigate)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,7 +235,7 @@ version = "0.28.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "829d955a0bb380ef178a640b91779e3987da38c9aea133b20614cfed8cdea9c6"
 dependencies = [
- "bitflags",
+ "bitflags 2.10.0",
  "crossterm_winapi",
  "mio",
  "parking_lot",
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -74,12 +74,34 @@ pub enum ConfigLoadError {
...
+    },
+}
+
 #[derive(Clone, Debug, Deserialize)]
 #[cfg_attr(feature = "json-schema", derive(schemars::JsonSchema))]
 #[serde(deny_unknown_fields)]
 pub struct DefaultsConfig {
     /// The theme to use by default in every presentation unless overridden.
-    pub theme: Option<String>,
+    pub theme: ThemeConfig,
```
mfontanini: This needs a `serde(default)` otherwise on a config file that has `defaults` but doesn't have `theme` it fails.

mfontanini (commented):
Nice, thanks for this!

PR #818, opened 2026-01-09
feat: use `icy_sixel` crate

gcavelier:
While trying to find the root cause of a bug when using presenterm on Windows Terminal (running WSL), I found the crate `icy_sixel` which implements the sixel encoding in pure Rust.

As the code is a little simpler and it doesn't rely on the C library provided by `sixel-sys`, I thought I could propose this PR ;)

PS: this did not fix my issue but I found the root cause to be in crossterm, see this [issue](https://github.com/crossterm-rs/crossterm/issues/1021)
PS2: I also updated the rust toolchain to its latest version

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -113,10 +113,42 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "812e12b5285cc515a9c72a5c1d3b6d46a19dac5acfef5265968c166106e31dd3"
 
 [[package]]
+name = "bitvec"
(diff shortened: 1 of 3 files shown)


mfontanini:
Hey, thanks! I had seen icy_sixel a while back and considered moving to it but I thought at the time it was a bit too immature or maybe not used widely enough. Based on your tests, does it work generally okay for the images you've tried?

I haven't had much time lately but I'll try to run this in the next few days and see how it works for me.

PR #839, opened 2026-02-04
feat: make typst background transparent

BenFradet:
## What

support alpha channels for typst colours

## Why

At the moment, if the theme you're using specifies typst colours, there is no way to go back to a transparent background.

### DIFF
--- a/src/markdown/text_style.rs
+++ b/src/markdown/text_style.rs
@@ -304,6 +304,7 @@ pub(crate) enum Color {
     White,
     Grey,
     Rgb { r: u8, g: u8, b: u8 },
+    Rgba { r: u8, g: u8, b: u8, a: u8 },
 }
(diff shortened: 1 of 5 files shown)


src/markdown/text_style.rs
```diff
@@ -304,6 +304,7 @@ pub(crate) enum Color {
     White,
     Grey,
     Rgb { r: u8, g: u8, b: u8 },
+    Rgba { r: u8, g: u8, b: u8, a: u8 },
```
mfontanini: I don't particularly like this approach since now any color can have an alpha component, but this has no effect anywhere except in the `typst` case. I think a better approach is to introduce a tiny `RgbaColor` type that's parseable from hex and use that inside the typst style config for its background color.

PR #845, opened 2026-02-16
Add styling docs to `slide_title`

0atman:
Out of interest, I tried to style the title, and it worked! I've updated the docs to reflect this.

Interestingly, I couldn't get `italic: true` to work, though bold and underline were fine.

### DIFF
--- a/docs/src/features/themes/definition.md
+++ b/docs/src/features/themes/definition.md
@@ -219,7 +219,7 @@ footer:
 
 ## Slide title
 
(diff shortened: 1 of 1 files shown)


0atman:
It finally happened, I started actually doing some work around here :laughing:

mfontanini:
Thanks!

mfontanini:
> Interestingly, I couldn't get italic: true to work, though bold and underline were fine.

It's _italics_, not _italic_. Will fix the docs!

Issue #863, opened 2026-03-14
Feature: Background Images


MarkusSagen:
In presentation software s.a. Powerpoint or Google Presentations it is possible to have background images and templates

Would be great to add something similar to this for branding your presentations
Mostly I would like this so that I can continue to use presenterm at work where they have certain templates to follow

0atman:
a nasty hack: misuse `jump-to-middle`:

```md
![](shai-hulhud.png)
<!-- jump_to_middle -->
# THOU SHALT NOT MAKE A MACHINE 
# IN THE LIKENESS OF A HUMAN MIND
```

<img width="1915" height="1065" alt="Image" src="https://github.com/user-attachments/assets/7242c952-b440-4678-a13c-92a8aebbb957" />

PR #864, opened 2026-03-14
feat: add background image support for slides

MarkusSagen:
* Support theme-level and per-slide bg images with: (cover, contain, stretch)
* can be set globally via theme config or overridden per-slide using
* New comment: 

Solves: #863

---

How it looks in terminal like Ghostty or Kitty

```rust
cargo run -- examples/background-image.md
```
Screenshot 2026-03-14 at 15 32 07

---

```rust
cargo run -- examples/background-image-cover.md
```
Screenshot 2026-03-14 at 15 32 13

---

```rust
cargo run -- examples/background-image-contain.md
```
Screenshot 2026-03-14 at 15 32 20

---


0atman:
VERY cool!

PR #875, opened 2026-04-07
feat: add mobile device support to HTML export

a-h:
I noticed that if you open a HTML slide export on a mobile device, you can't progress the slide content, because there's no tap/touch and swipe support.

I've updated the script to do this, and also updated the export code to not require a TTY to run, choosing some defaults targeted towards a 13" laptop display, but that work well in landscape mode on mobile devices.

I also fixed a bug where you might get an error like this in the browser console when trying to load the slide content from a file on disk instead of from a web server:

```
presenterm-demo.html#slide-2:1 Unsafe attempt to load URL file:///private/tmp/presenterm-demo.html#slide-2 from frame with URL file:///private/tmp/presenterm-demo.html#slide-2. 'file:' URLs are treated as unique security origins.
```

1600x900

iPhone 12 Pro

iphone_12_pro

Pixel 7

pixel_7

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -457,16 +457,7 @@ fn run(cli: Cli) -> Result<(), Box<dyn std::error::Error>> {
                 height: dimensions.rows * DEFAULT_EXPORT_PIXELS_PER_ROW,
                 width: dimensions.columns * DEFAULT_EXPORT_PIXELS_PER_COLUMN,
             },
-            None => {
-                const DEFAULT_EXPORT_ROWS: u16 = 40;
-                const DEFAULT_EXPORT_COLUMNS: u16 = 120;
-                WindowSize {
-                    rows: DEFAULT_EXPORT_ROWS,
-                    columns: DEFAULT_EXPORT_COLUMNS,
-                    height: DEFAULT_EXPORT_ROWS * DEFAULT_EXPORT_PIXELS_PER_ROW,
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -457,7 +457,16 @@ fn run(cli: Cli) -> Result<(), Box<dyn std::error::Error>> {
                 height: dimensions.rows * DEFAULT_EXPORT_PIXELS_PER_ROW,
                 width: dimensions.columns * DEFAULT_EXPORT_PIXELS_PER_COLUMN,
             },
-            None => WindowSize::current(config.defaults.terminal_font_size)?,
+            None => {
```
mfontanini: I would rather not do this. If you really don't want to use a TTY you should use the configurable dimensions via `export.dimensions`. This is very arbitrary and will look ok for some but not for everyone so I'd rather not make assumptions.

mfontanini:
This is great, thanks! I would just not do the dimensions thing. If you can take that I can merge right away. Sorry it took a while to get to this, I've had 0 time lately.

src/main.rs
a-h: Thanks, I've made the changes.

a-h:
Thanks, no need to say sorry. I appreciate the tool, and your work to maintain it.

I've been a long time fan of markdown presentation tools, but I especially appreciate that this is a single binary without loads of boilerplate around it, and the ease of using d2 / typst.

mfontanini:
Thanks!

PR #887, opened 2026-04-30
feat: add exec support for scala 3

hoop33:
Adds support for executing Scala code snippets. Note that it works with Scala 3, not Scala 2.

I spent some time trying to add an alternative executor for Scala 2, doing a `scalac` to compile and a `scala` to execute. I could get it to work if the class in the snippet was called `Snippet`, which seems like an esoteric restriction. So I gave up on Scala 2.


mfontanini:
Thanks!

